Issue #119, opened 2026-08-04
chore: redo thumbnails for Wordleverse and Hashtag


csalinas-dev:
Update the thumbnails on the Wordleverse and Hashtag game tiles. Also, update the layout for the game tiles. Make it full width instead of in a container

---

## 🤖 Implementation Plan

_Written by the architect agent on 2026-08-07 08:28 UTC. The implementor builds exactly this._

## 🤖 Implementation Plan

_Re-planned on 2026-08-07 after PR #122 (issue #120) landed. The previous plan
targeted the pre-refactor 443-line `games/page.jsx` and is superseded in full —
ignore any file reference you remember; everything below is written against
`main` at commit `43afd31`._

### Goal

`/games` stops mixing two phone screenshots with three illustrations: Wordleverse
and Hashtag get hand-drawn inline-SVG artwork in the same house style as
Tic-Tac-Overflow, Edge Case and Connect 404, and the tile grid runs the full width
of the viewport instead of being capped by the MUI `xl` container. Two orphaned
JPGs (714 KB) leave the repo.

### Approach

**Thumbnails: draw them, don't re-shoot them.** #120 gave every game its own
artwork module under `src/app/(pages)/games/_artwork/`. Three of them are
hand-drawn SVG; `Wordleverse.jsx` and `Hashtag.jsx` are still `<Image fill>` over a
tall phone screenshot cropped `object-position: top`, so they show mostly app
chrome and read as a different kind of thing to the other three. Replace the body
of those two modules with a 300×300 SVG drawing in the shared VS Code palette, and
delete the JPGs. The rejected alternative — retaking the screenshots at the card
ratio — loses because it leaves the tiles stylistically inconsistent, keeps ~714 KB
of binaries in the repo, and needs a manual capture step nobody can reproduce.

**Layout: full-bleed, gutters equal to the grid gap.** `page.jsx` still wraps the
grid in `<Section>`, which is `Container maxWidth="xl"` — #120 explicitly left this
alone (it is in that issue's out-of-scope list), so the user's "make it full width
instead of in a container" is still entirely undone. Replace `<Section>` with a
page-local full-bleed section and move the horizontal gutter onto `GameGrid`, next
to the `gap` it must match. The cards do *not* run to a zero gutter: `Card` has
`border-radius: 1rem` plus a box-shadow, and `globals.css` sets `overflow-x: hidden`
on `body`, so a zero-gutter edge clips both. A gutter equal to the grid gap makes
[...]

PR #122, opened 2026-08-07
refactor(games): one card component, one grid, one manifest for /games

csalinas-dev:
Closes #120

## What changed

- **`games.js` is the manifest.** One entry per game (`{ slug, title, href, Artwork }`) in display order, `Artwork` being the component reference so the file stays plain JS. The Mini Motorways comment moved here — it explains that `/games/mini-motorways` still exists and this list is only what is *linked*.
- **`_components/GameCard.jsx` is the whole tile** (`"use client"`, since it uses Emotion). `Card` keeps the old radius and three-layer shadow verbatim and adds `aspect-ratio: 1 / 2`, `container-type: inline-size` and one hover rule (`&:hover .media`). `Media` keeps the radial gradient, the 8% inset and the 250ms scale, and its SVG rule collapses to a single `width: 100%` form because the card is now always portrait. `img { object-fit: cover; object-position: top; }` preserves the photo tiles' crop now that they are no longer a `styled(Image)`.
- **The title is sized in container units.** `font-size: clamp(0.875rem, 13cqw, 2.75rem)` and `padding: 10cqw 7cqw` replace five viewport media queries. Those queries assumed card width tracks viewport width, which stops being true once the column count is fluid.
- **`GameGrid` is one `repeat(auto-fit, minmax(min(9rem, 100%), 1fr))` rule** exported from the same file. No game is named in any CSS; empty tracks collapse and the survivors share the row.
- **`_artwork/` is one module per game.** `TicTacOverflow.jsx`, `EdgeCase.jsx` and `Connect404.jsx` are the existing SVGs moved byte-for-byte, comments included. `Wordleverse.jsx` and `Hashtag.jsx` wrap the existing `@/assets/*.jpg` in `<Image fill placeholder="blur">` with `alt=""` (decorative — the card's own text already says "Play Wordleverse"). Those two files are the seam #119 replaces.
- **`page.jsx` is 19 lines** and no longer declares `"use client"`: nothing left in it is client-only, so `/games` is a server component. It still prerenders static (`○ /games` in the build output).
- **`CLAUDE.md`** gains one bullet under Important Patterns describing the manifest, the card-owns-the-aspect-ratio rule and the square `viewBox="0 0 300 300"` artwork convention.

### Adding a sixth game

Per the acceptance criteria, the exact diff a Mini Motorways tile would need — two files, no CSS:

1. New `src/app/(pages)/games/_artwork/MiniMotorways.jsx`: an `import Image from "next/image"` + `import miniMotorways from "@/assets/mini-motorways.jpg"` and a `const MiniMotorwaysArtwork = () => <Image alt="" fill placeholder="blur" src={miniMotorways} />; export default MiniMotorwaysArtwork;` (the asset already exists).
2. In `src/app/(pages)/games/games.js`: one `import MiniMotorwaysArtwork from "./_artwork/MiniMotorways";` and one array entry `{ slug: "mini-motorways", title: "Mini Motorways", href: "/games/mini-motorways", Artwork: MiniMotorwaysArtwork }`, minus the Mini-Motorways-is-hidden paragraph of the comment.

Nothing in `GameCard.jsx` or `page.jsx` changes. (Not added here — Mini Motorways stays unlinked, per the plan.)

## How it was verified

    npm install
    npx prisma generate      # the worktree is a fresh clone; build needs the client
    npm run lint
    ✔ No ESLint warnings or errors

    npm run build
    ✓ Compiled successfully in 29.8s
    ○ /games      6.6 kB      159 kB      (still prerendered as static content)

    npx next start -p 3120 && curl http://localhost:3120/games

Against the served markup: `aspect-ratio:1/2` and `container-type:inline-size` on the card, `aspect-ratio:1/1` on the SVG, `grid-template-columns:repeat(auto-fit, minmax(min(9rem, 100%), 1fr))`, `font-size:clamp(0.875rem, 13cqw, 2.75rem)`, `padding:10cqw 7cqw`, `object-position:top`; no `grid-template-areas` anywhere; three `viewBox="0 0 300 300"` SVGs and two `alt=""` images; all five hrefs present and unchanged (`/games/wordleverse`, `/games/hashtag`, `/games/tic-tac-overflow`, `/games/edge-case`, `/games/connect-404`).

## Notes for the reviewer

- **I could not drive a browser from this agent context**, so the responsive checks are the emitted CSS plus arithmetic rather than screenshots. The grid maths, with `Section`'s confirmed `padding: 2rem` (available width = viewport − 64):
  | viewport | gap | columns | card width |
  |---|---|---|---|
  | 320 | 1rem | 1 | 256px (no overflow — `min(9rem, 100%)`) |
  | 375 | 1rem | **2** | 147px |
  | 768 | 1rem | 4 | 164px |
  | 900 | 2rem | 4 | 185px |
  | 1296 | 2rem | **5** | 221px |
  | 1440 | 2rem | **5** | 250px |
  Worth a human eyeballing before merge.
- **Two acceptance-list items deserve a flag.**
  1. The criterion reads "at 1440px (5 narrow columns) the title is visibly smaller than at 900px (2 wide columns)". With the `9rem` min track the plan specifies, 900px yields **4** columns of 185px, not 2 wide ones — so the title is *larger* at 1440 (250px card) than at 900 (185px card). The underlying intent, that the title tracks the card and not the window, does hold; the parenthetical column counts in the criterion just don't match the grid rule the same plan specifies. I built the rule as written rather than retuning `9rem` to make the parenthetical true.
  2. Consequently five cards wrap **4 + 1** in roughly 688-848px and 896-912px, leaving an orphan on the second row at common tablet widths. That is inherent to `auto-fit` with a fixed min track and an odd card count; lowering the min track would fix 768px but risks the 375px-shows-2-columns criterion. Say the word and I'll retune.
- **The hover now scales the whole media frame, not just the image.** Previously `&:hover img, &:hover .artwork` scaled the `<Image>` itself on photo tiles and the artwork div on SVG tiles. Since the image fills the frame, the result is visually the same, but it is a real change in which element carries the transform.
- **`<Image fill>` still has no `sizes` prop**, so Next will keep logging its dev-mode warning for the two photo tiles. That is pre-existing (the old `CardImage` had the same), and adding `sizes` would be a behaviour change #119 may want to make with its new thumbnails, so I left it.
- **`mask id="connect404-holes"` is still a document-global id**, moved unchanged. Fine at one card per page; noted in the plan as a pattern not to copy.
- The duplicate game list in `src/components/Nav.jsx` is untouched, per Out of scope.
- Rebased on `origin/main` (`204304a`) immediately before pushing; #119 has not merged yet, so there was nothing to reconcile.


csalinas-dev (commented):
## 🤖 Review: APPROVED

_(Posted as a comment — GitHub will not let the same account formally approve its own PR.)_

Checked: security, correctness, efficiency, fit against the plan. This is a pure refactor, so the bar was "nothing changed for the three artwork games" — I verified that mechanically, and I drove a real browser rather than inheriting the arithmetic in the PR body.

**Verified**

- **The three SVGs really did move byte-for-byte.** Extracted lines 157-210 / 212-285 / 287-367 from `page.jsx` at `204304a` and diffed them against the new `_artwork/*.jsx` minus each file's trailing `export default`: all three `IDENTICAL`, comments included.
- `npm install` → `prisma generate` → `npm run lint` (✔ no warnings) → `npm run build` (✓ compiled, `○ /games` still prerendered static, so dropping `"use client"` held).
- **All five hrefs resolve**: `/games/{wordleverse,hashtag,tic-tac-overflow,edge-case,connect-404}` all return 200 off `next start`. Emitted markup has 3× `viewBox="0 0 300 300"`, 2× `alt=""`, zero `grid-template-areas`.
- **Real browser, headless Chrome 151 over CDP**, measured at 320/375/414/600/624/640/688/768/800/840/848/880/890/896/900/912/1024/1296/1440/1920:
  - `height / width` is exactly `2.000` on every card at every width — no card is landscape anywhere.
  - No horizontal overflow at any width, including 320px.
  - 375px → 2 columns (148px cards). 1440px → 5 equal columns of 253px in one row. Both pinned criteria pass.
  - Computed style on the photo tiles is `object-fit: cover; object-position: 50% 0%`, and the `` box equals the full card box (253×506) — the `fill` image correctly ignores `Media`'s 8% padding, and the Emotion `img` selector is matching. The crop is preserved.
  - SVG tiles measure 212×212 inside a 253px card (253 − 2×8%), square as intended.
- **Hover**: dispatched real `Input.dispatchMouseEvent` moves onto a photo tile and an SVG tile — `.media` computes to `matrix(1.05, 0, 0, 1.05, 0, 0)` on both, title stays `none`. The transform moved to a different element but the behaviour is identical.
- **Console clean**: no `react/jsx-key` warnings, no hydration errors. (Only `next-auth` 500s from missing env in my sandbox, unrelated to this PR.)
- **Accessibility tree**: all five links have names — `Play Wordleverse`, `Play Hashtag`, `Play Tic-Tac-Overflow`, `Play Edge Case`, `Play Connect 404`. Titles wrap at the hyphen (`Tic-Tac-` / `Overflow`) and at the space (`Connect` / `404`), never mid-word, and `overflowsX` is false on every card at every width.
- `grep` for `CardArtwork|CardImage|tictacoverflow|.artwork|grid-template-areas` across `src/` finds nothing but the unrelated Wordleverse game board. `Nav.jsx` and both jpgs untouched; diff scope matches the plan's Changes list exactly.

**Acceptance criteria**

- [x] No SVG / `grid-template-areas` / per-game CSS left in `page.jsx`; one `.map` over `GAMES`.
- [x] `1 / 2` at every width — measured 2.000 at 20 widths.
- [x] One chrome: radius, three-layer shadow, gradient overlay and hover-scale are all in `GameCard` and shared by all five.
- [x] 375px → 2 columns; 1440px → 5 equal columns in one row; no h-scroll at 320px.
- [x] Title scales with the card, not the window — see adjudication below.
- [x] All five hrefs resolve (200).
- [x] Sixth-game diff described in the PR body (two files, no CSS).
- [x] `npm run lint` clean.
- [x] `page.jsx` has no `"use client"` and `/games` still prerenders.

---

### Adjudicating the two self-reported flags

**1. "At 900px, 2 wide columns" — the criterion is wrong, not the build. Not blocking.**

I did not take the implementor's arithmetic on faith; I swept it live in the browser, holding the grid rule and varying only the min track:

| min track | 375px | 900px | 1440px |
|---|---|---|---|
| 6–9rem | **2** | 5 | **5** |
| 10rem | 1 | 4 | 5 |
| 12–14rem | 1 | 3 | 5 |
| 16rem | 1 | 3 | 4 |
| 18–20rem | 1 | **2** | 4 |
| 24rem | 1 | **2** | 3 |

No value satisfies all three. Getting 2 columns at 900px requires ≥18rem, which costs you *both* of the criteria the plan itself names as "the two that pin it down" (375px→1 column, 1440px→4 columns). The criterion's `(2 wide columns)` parenthetical is unsatisfiable alongside the rule the same plan mandates. Building the rule as written was the right call.

Note the PR body's own arithmetic is slightly off too: 900px is **5** columns of 145px, not 4 of 185px.

The criterion's *substance* — title tracks the card, not the window — is met, and here is the demonstration it was reaching for: **880px viewport → card 154px, title 19.97px; 896px viewport → card 144px, title 18.72px.** The window gets 16px wider and the title gets smaller, because the card did. The old viewport media queries would have jumped that same 144px card to `2.75rem`.

**2. The 4+1 orphan ships. Non-blocking — but the reported band is wrong.**

Measured, the orphan band is **~656–839px only**. 848 / 880 / 890 / 894 / 896 / 898 / 904 / 912 all render **5 in a single row** — the claimed second band at 896–912px does not exist (at 896 the gap goes to 2rem and 5×144 + 4×32 = 848 fits the grid width exactly). Column count is monotone in viewport width; there is no widen-and-lose-a-column flip. Worth correcting on the record.

The remaining 656–839px orphan ships, because:
- Five items in a fluid grid orphan at *some* width by definition — the criteria already bless 2+2+1 at 375px, which is the same phenomenon.
- No acceptance criterion pins any width inside the band.
- Removing it means naming "five" in CSS, which is precisely what #120 exists to delete and what the plan's rejected-alternative section rules out.

**Non-blocking observations**

- The blur placeholder's inline style is `background-position: 50% 50%` while the loaded photo is `object-position: top`, so the two photo tiles shift slightly as they decode. Pre-existing — the old `styled(Image)` split them the same way — and #119 is replacing exactly those two files, so fold it in there.
- `<Image fill>` with no `sizes` resolves to `sizes="100vw"`, so a 1080px source is fetched for a 253px card. Also pre-existing; also #119's territory.
- At 320px the grid drops to one column, giving five 256×512 tiles and ~2.7k px of scroll. Within the criteria (only "no h-scroll at 320px" is pinned) but it is the one width where the uniform portrait shape costs real scrolling.

PR #128, opened 2026-08-07
fix(wordleverse): make the opt-in failure message tell the truth in both directions

### LINKED ISSUE #126: fix(wordleverse): opt-in failure message asserts the wrong listing state
The Wordleverse leaderboard opt-in toggle shows one shared failure message for every
failure path:

`src/app/(pages)/games/wordleverse/leaderboard/_components/OptInToggle.jsx:23`

```js
const FAILED =
  "Failed to update leaderboard setting. You are still listed as before.";
```

That sentence is correct for a failed opt-**out** and wrong for a failed opt-**in**.
A user who is *not* on the leaderboard, clicks the switch ON, and hits an error is
told they are "still listed" — asserting their name is published when it is not.

This was found during review of #14 (PR #123) and approved as non-blocking, because
it errs toward over-warning rather than under-warning, so it is not a privacy fault.
But it states a falsehood about publication status inside a feature whose entire
premise is consent, so it should not stay.

The fix is wording, not logic — the failure handling itself was already fixed and
verified in that review (a thrown action leaves a listed user's switch ON with a
visible alert). Do not change the fail-closed behaviour; only make the message tell
the truth in both directions.

Note also that the two failure modes are already distinguishable: a rejection shows
`FAILED`, while a returned `{ error }` shows the server action's own string. So the
message can be made direction-aware without new plumbing.

### Done when
- A user whose opt-**in** fails is not told they are listed.
- A user whose opt-**out** fails is still clearly told they remain listed.
- The fail-closed behaviour is unchanged: any failure leaves the switch showing the
  user's real, server-side state, never an optimistic one.

---

## 🤖 Implementation Plan

_Written by the architect agent on 2026-08-07 16:25 UTC. The implementor builds exactly this._

### Goal

When the Wordleverse leaderboard opt-in switch fails, the error message states the
listing status that is actually true — "you have not been added" after a failed
opt-**in**, "you are still listed" after a failed opt-**out** — instead of one
sentence that only happens to be true in the opt-out direction.

This is a **wording-only** change. No control flow, no state handling, no server
action logic changes.

### Approach

`onChange` already knows the direction: `next` is the position the user asked for,
and every failure path reverts the switch to `!next`. So the message can be made
direction-aware by computing one clause from `next` at the top of the handler and
appending it to whatever reason is being reported. No new plumbing, no new state.

Both failure paths get the clause, not just the `catch`. The `{ error }` return is
the *more* common failure (the server action catches everything and returns rather
than throwing — see `setLeaderboardOptIn.js:38-41`), and today that path renders the
bare server string with no statement about listing at all. Fixing only the `FAILED`
constant would leave the issue's second Done-when ("a failed opt-out is still
clearly told they remain listed") unmet on the path most users would hit.

**Rejected:** hedging the opt-in message ("reload to check whether you were added")
to cover the rare case where the write committed but the response was lost. Under
the fail-closed contract verified in PR #123, the switch itself asserts `!next` —
prose that hedges while the switch sits confidently OFF is incoherent. Both
sentences should agree with the switch, which is the last state the server
confirmed.

- The only file to change is
  `src/app/(pages)/games/wordleverse/leaderboard/_components/OptInToggle.jsx`.
  `FAILED` is declared at line 23 and used at **two** sites: line 66 (the
  `typeof result?.optIn !== "boolean"` branch, as a fallback behind
  `result?.error`) and line 80 (the `catch`).
- `UNKNOWN` (line 25) is a different message for a different condition — the
  setting could not be read at all. **Do not touch it.**
- The server action `_actions/setLeaderboardOptIn.js` returns exactly two error
  strings, `"Unauthorized"` and `"Failed to update leaderboard setting"`, and
  **neither ends in punctuation**. That is why the template below supplies the
  period. Do not add periods to the action's strings.
- The fail-closed mechanism (`setChecked(!next)` on both failure paths, the
  `getCurrentUser()`-inside-`try`, `optedIn: null` disabling the switch) was fixed
  and independently verified in the review of PR #123 — a reviewer mounted the
  component in jsdom and asserted 40 failure shapes. **Leave all of it alone.**
- This repo has no test runner: `package.json` scripts are only
  `dev` / `build` / `start` / `lint`, and `.github/workflows/deploy.yml` runs no
  tests. Do not add a test framework for this.

### Changes

1. **`src/app/(pages)/games/wordleverse/leaderboard/_components/OptInToggle.jsx`**

   Replace the `FAILED` constant (line 23-24) with a reason prefix plus the two
   direction clauses:

   ```js
   const FAILED = "Failed to update leaderboard setting";
   // Every failure path puts the switch back where it was, so what is true
   // afterwards is the state the click tried to leave. One shared sentence
   // cannot say that: telling somebody whose opt-in failed that they are "still
   // listed" asserts a publication that never happened.
   const STILL_LISTED = "You are still listed on the leaderboard.";
   const NOT_LISTED = "You have not been added to the leaderboard.";
   ```

   In `onChange`, immediately after `const next = event.target.checked;`:

   ```js
   // Both failure paths below revert to `!next`, so this is what is true then.
   const unchanged = next ? NOT_LISTED : STILL_LISTED;
   ```

   Then at the two failure sites:

   ```js
   // was: setError(result?.error ?? FAILED);
   setError(`${result?.error ?? FAILED}. ${unchanged}`);
   ```

   ```js
   // was: setError(FAILED);
   setError(`${FAILED}. ${unchanged}`);
   ```

   The four resulting messages:

   | attempt | failure | message |
   |---|---|---|
   | opt-in | thrown | `Failed to update leaderboard setting. You have not been added to the leaderboard.` |
   | opt-in | `{ error: "Unauthorized" }` | `Unauthorized. You have not been added to the leaderboard.` |
   | opt-out | thrown | `Failed to update leaderboard setting. You are still listed on the leaderboard.` |
   | opt-out | `{ error: "Failed to update…" }` | `Failed to update leaderboard setting. You are still listed on the leaderboard.` |

2. **`src/app/(pages)/games/wordleverse/CLAUDE.md`** — the leaderboard bullet that
   starts "**The opt-in switch is optimistic…**" (around line 108) already
   documents the revert-and-report rule. Append one sentence to that same bullet:
   the failure message is direction-aware because the switch reverts to the state
   the click tried to leave, so "still listed" belongs only to a failed opt-out.
   Do not restructure the bullet or add a new one.

### Out of scope

- `setLeaderboardOptIn.js` — its strings, its `try` boundary, its return shape.
- Any change to `setChecked`, `saving`, `pending`, `unknown`, or the `disabled`
  logic. The fail-closed behaviour is verified and settled; if you believe it is
  wrong, stop and say so on the issue rather than changing it.
- `UNKNOWN`, `getLeaderboard()`'s `optedIn: null` path, `rank.js`, the boards.
- Adding a test runner, jsdom setup, or CI test step.

### Acceptance criteria

- [ ] A failed opt-in (switch was OFF, clicked ON, request fails) shows a message
      that does **not** contain the word "listed" as an assertion about the user —
      it says they have not been added.
- [ ] A failed opt-out (switch was ON, clicked OFF, request fails) shows a message
      that explicitly says they are still listed.
- [ ] Both failure paths carry the clause: the `catch` **and** the branch where the
      action returns `{ error }`. An `{ error: "Unauthorized" }` on a failed opt-out
      still tells the user they remain listed.
- [ ] The switch position on failure is unchanged from today: it reverts to the
      pre-click, server-confirmed state on every failure path.
- [ ] `UNKNOWN` and the `optedIn == null` disabled state are byte-identical to main.
- [ ] The diff touches only `OptInToggle.jsx` and the one `CLAUDE.md` bullet.

### Verification

    npm run lint

There is no test script in this repo. Manually confirm the two directions by
temporarily forcing a failure in `onChange` (e.g. `throw new Error("x")` before the
`await`), loading `/games/wordleverse/leaderboard` signed in, toggling in each
direction, and reading the `role="alert"` text — then **revert the temporary
throw** before committing.

Optionally, `npm run build` also exercises lint/type checking, but it needs a
reachable `DATABASE_URL` for Prisma generate; `npm run lint` is the required gate.

### Risks

- **Double punctuation.** If a future error string ever ends in a period, the
  template produces `"… setting.. You are …"`. Both current strings are
  unpunctuated; the comment in the code should not need to restate this, but do not
  add trailing periods to `setLeaderboardOptIn.js`.
- **Missing the second call site.** The `FAILED` fallback at line 66 is easy to
  overlook because it sits behind `result?.error ??`. Grep for `FAILED` after
  editing and confirm two usages, both with the appended clause.

csalinas-dev:
Closes #126

## What changed

Wording only, in `src/app/(pages)/games/wordleverse/leaderboard/_components/OptInToggle.jsx`.

- `FAILED` loses its trailing listing claim and becomes just the reason,
  `"Failed to update leaderboard setting"` (no period — the two strings
  `setLeaderboardOptIn` returns are unpunctuated too, so the template supplies it).
- Two new direction clauses, `STILL_LISTED` / `NOT_LISTED`.
- `onChange` computes `const unchanged = next ? NOT_LISTED : STILL_LISTED;`
  immediately after `next`. Both failure paths revert to `!next`, so `unchanged`
  is by construction what is true once the revert has happened.
- **Both** failure sites append it, not just the `catch`. The
  `typeof result?.optIn !== "boolean"` branch (the `result?.error ?? FAILED`
  fallback) is the path most users actually hit — the server action catches and
  returns rather than throwing — and it previously said nothing about listing at
  all, so fixing only the constant would have left the issue's second Done-when
  unmet on the common path.
- `src/app/(pages)/games/wordleverse/CLAUDE.md`: one sentence appended to the
  existing "The opt-in switch is optimistic…" bullet. No new bullet, no
  restructuring.

Nothing else moved. `UNKNOWN`, `setChecked`, `saving`, `pending`, the `disabled`
logic, the `optedIn == null` path and `setLeaderboardOptIn.js` are byte-identical
to `main` — the fail-closed mechanism verified in the review of #123 was left
alone deliberately.

## How it was verified

    npm ci
    npm run lint
    → ✔ No ESLint warnings or errors

Lint is the only gate this repo has (`package.json` scripts are `dev` / `build` /
`start` / `lint`; `deploy.yml` runs no tests) and the plan explicitly ruled out
adding a test runner for a wording fix. **Lint passing says nothing about the
message text.** So I separately drove the handler, in a throwaway harness that is
not part of this PR:

it loads the real `OptInToggle.jsx` source, transpiles the JSX with the repo's own
`typescript`, stubs the module's imports (React's `useState`/`useTransition`, MUI,
emotion, `next/navigation`, and the server action), pulls the real `onChange` off
the returned element, calls it, and reads back the `error` string and `checked`
value the component would render.

| case | switch after | message |
|---|---|---|
| opt-in, action throws | OFF (was OFF) | `Failed to update leaderboard setting. You have not been added to the leaderboard.` |
| opt-in, `{ error: "Unauthorized" }` | OFF (was OFF) | `Unauthorized. You have not been added to the leaderboard.` |
| opt-in, unrecognised shape `{}` | OFF (was OFF) | `Failed to update leaderboard setting. You have not been added to the leaderboard.` |
| opt-out, action throws | ON (was ON) | `Failed to update leaderboard setting. You are still listed on the leaderboard.` |
| opt-out, `{ error: "Failed to update leaderboard setting" }` | ON (was ON) | `Failed to update leaderboard setting. You are still listed on the leaderboard.` |
| opt-out, `{ error: "Unauthorized" }` | ON (was ON) | `Unauthorized. You are still listed on the leaderboard.` |
| opt-in succeeds (control) | ON (was OFF) | no alert |

No double punctuation in any case, and the switch reverts to the pre-click state
on every failure and sticks only on the confirmed success.

## Notes for the reviewer

Things I would flag at you before you find them:

- **What is still unverified.** The harness stubs React's hooks and MUI, so it
  exercises the handler's logic and the exact strings it sets — it does **not**
  prove React re-renders, that the string reaches the `role="alert"` span, or
  anything about the rendered DOM. I did not load the page in a browser: that
  needs a signed-in session against a live `DATABASE_URL`, which I do not have
  here. The render path is unchanged from `main` (`{error && <ErrorMessage
  role="alert">{error}</ErrorMessage>}`), so I judged the risk there to be nil,
  but I am not claiming to have observed it.
- **Double punctuation is a live hazard for future strings.** The template is
  `` `${reason}. ${unchanged}` ``. Both strings `setLeaderboardOptIn` returns today
  are unpunctuated, so this is correct now; a future error string ending in a
  period would render `"… setting.. You are …"`. I did not add a normalising
  strip, because stripping trailing punctuation from an arbitrary server string
  is more machinery than a two-string call site earns, and the constraint lives
  next to both strings. Say the word if you would rather have the strip.
- **I did not hedge the opt-in message.** If the write commits but the response is
  lost, the user is told they were not added while they actually were. That is the
  same window the pre-existing fail-closed contract already accepts — the switch
  itself asserts `!next` — so hedging the prose while the switch sits confidently
  OFF would be incoherent. This was considered and rejected in the plan; flagging
  it because it is the one case where the new sentence can be wrong.
- **Direction of the error.** For a failed opt-in the message now under-warns
  relative to the old one (it says "not added" rather than "still listed"). That
  is the point of the issue, but it is worth stating explicitly: if you think the
  privacy-safe default is to over-warn in both directions, that is a design
  disagreement with the issue, not with this diff.

### DIFF
--- a/src/app/(pages)/games/wordleverse/CLAUDE.md
+++ b/src/app/(pages)/games/wordleverse/CLAUDE.md
@@ -115,7 +115,11 @@ Prisma read, with no client fetch.
   on a returned error and in a `catch`. Only a boolean the server confirmed
   survives. For the same reason `getLeaderboard()`'s catch reports `optedIn:
   null`, not `false`: the setting was never read, and "you are not listed" would
-  be a guess. `null` disables the switch and suppresses the notice.
+  be a guess. `null` disables the switch and suppresses the notice. The failure
+  message is direction-aware for the same reason — the revert leaves the user in
+  the state the click tried to leave, so "you are still listed" belongs only to a
+  failed opt-*out*, and both failure paths append the clause, not just the
+  `catch`.
 
 ## Key Invariants
 
--- a/src/app/(pages)/games/wordleverse/leaderboard/_components/OptInToggle.jsx
+++ b/src/app/(pages)/games/wordleverse/leaderboard/_components/OptInToggle.jsx
@@ -20,8 +20,13 @@ const ErrorMessage = styled.span`
   font-size: 0.9rem;
 `;
 
-const FAILED =
-  "Failed to update leaderboard setting. You are still listed as before.";
+const FAILED = "Failed to update leaderboard setting";
+// Every failure path puts the switch back where it was, so what is true
+// afterwards is the state the click tried to leave. One shared sentence cannot
+// say that: telling somebody whose opt-in failed that they are "still listed"
+// asserts a publication that never happened.
+const STILL_LISTED = "You are still listed on the leaderboard.";
+const NOT_LISTED = "You have not been added to the leaderboard.";
 const UNKNOWN =
   "Could not read your leaderboard setting. Reload to try again.";
 
@@ -50,6 +55,8 @@ const OptInToggle = ({ optedIn }) => {
 
   const onChange = async (event) => {
     const next = event.target.checked;
+    // Both failure paths below revert to `!next`, so this is what is true then.
+    const unchanged = next ? NOT_LISTED : STILL_LISTED;
 
     // Move the switch immediately, put it back unless the server confirms it.
     setChecked(next);
@@ -63,7 +70,7 @@ const OptInToggle = ({ optedIn }) => {
       // returned { error }, or a shape this component does not recognise.
       if (typeof result?.optIn !== "boolean") {
         setChecked(!next);
-        setError(result?.error ?? FAILED);
+        setError(`${result?.error ?? FAILED}. ${unchanged}`);
         return;
       }
 
@@ -77,7 +84,7 @@ const OptInToggle = ({ optedIn }) => {
       // they opted out.
       console.error("Error setting leaderboard opt-in:", error);
       setChecked(!next);
-      setError(FAILED);
+      setError(`${FAILED}. ${unchanged}`);
     } finally {
       setSaving(false);
     }


csalinas-dev:
## 🤖 Review: APPROVED

Posted as a **comment, not a formal approval** — GitHub refuses an approve/request-changes review on a self-authored PR, and every agent here runs as `csalinas-dev`. Treat this as the reviewer verdict; it is not a GitHub-recorded approval.

### What I did

I did not take the "byte-identical to main" claim on faith, and I did not take the implementor's harness on faith either — it stubbed React's hooks and MUI, so it observed the strings `onChange` *sets*, not the strings a user *sees*. I wrote my own harness that closes that gap: jsdom + **real** `react-dom/client`, the **real** `OptInToggle.jsx` (JSX transpiled with the repo's own `typescript`), the **real** MUI `Switch`. Only `next/navigation`'s `useRouter` and the server action are stubbed. Each case clicks the actual `<input type="checkbox">` and reads `textContent` off the rendered `[role="alert"]` node.

Then I ran the identical harness against `origin/main` and diffed the results case-for-case.

**20 cases. 0 mechanism differences.**

Switch position and `disabled` state, before and after, are identical to `main` in every one of: opt-in and opt-out × {sync throw, rejected promise, `{error:"Unauthorized"}`, `{error:"Failed to update leaderboard setting"}`, `{}`, `undefined`, `null`, `{optIn:"true"}`}, plus both success paths, plus the server-stored-the-opposite case, plus `optedIn: null` and `optedIn: undefined`. Every failure reverts to `!next`; only a confirmed boolean sticks; the unknown state stays disabled with `UNKNOWN` unchanged. The PR #123 fail-closed contract is intact — the claim holds, and it now holds by measurement rather than by inspection.

Every message difference vs `main` is an intended one. The four shapes that matter, read out of the live `role="alert"` span:

| attempt | failure | rendered alert | switch |
|---|---|---|---|
| opt-in | throws / rejects | `Failed to update leaderboard setting. You have not been added to the leaderboard.` | OFF |
| opt-in | `{error:"Unauthorized"}` | `Unauthorized. You have not been added to the leaderboard.` | OFF |
| opt-out | throws / rejects | `Failed to update leaderboard setting. You are still listed on the leaderboard.` | ON |
| opt-out | `{error:"Unauthorized"}` | `Unauthorized. You are still listed on the leaderboard.` | ON |

The returned-`{error}` path — the common one, since `setLeaderboardOptIn` catches and returns rather than throwing — is the one that changed most: on `main` it rendered a bare `"Unauthorized"` with **no** statement about listing at all. It now carries a direction-correct clause. That was the plan's judgement call and it was the right one.

`npm ci && npm run lint` → `✔ No ESLint warnings or errors`. `grep` confirms `FAILED` has exactly two call sites, both with the clause. Diff is 2 files / 3 hunks; `setLeaderboardOptIn.js` is not touched.

### Acceptance criteria

- [x] Failed opt-in does not assert listing — says "have not been added". Observed in the DOM, 8 opt-in failure shapes.
- [x] Failed opt-out still says "still listed" explicitly. Observed, 5 opt-out failure shapes.
- [x] **Both** paths carry the clause — `catch` and the `{ error }` branch. `{error:"Unauthorized"}` on a failed opt-out renders `Unauthorized. You are still listed on the leaderboard.`
- [x] Switch position on failure unchanged from today. 0/20 mechanism differences vs `origin/main`.
- [x] `UNKNOWN` and the `optedIn == null` disabled state byte-identical. Not in the diff; harness confirms `disabled: true` + `UNKNOWN` for both `null` and `undefined`.
- [x] Diff touches only `OptInToggle.jsx` and the one CLAUDE.md bullet. The bullet was appended to, not restructured.

### Adjudicating the four self-reported concerns

1. **Double punctuation — real, reproduced, correctly deprioritised.** I forced `{ error: "Nope." }` and got `Nope.. You have not been added to the leaderboard.` It is unreachable today: `setLeaderboardOptIn.js:27` and `:40` return `"Unauthorized"` and `"Failed to update leaderboard setting"`, both unpunctuated. I agree that stripping trailing punctuation off an arbitrary server string is more machinery than a two-string call site earns. The one thing I would change: the constraint currently lives in the plan and the PR body, and neither is where the next person editing that action will look. Not blocking.
2. **Commit-succeeded-response-lost.** Agree with the plan's rejection. The switch already asserts `!next` in that window and has since #123; prose that hedges while the switch sits confidently OFF would contradict the control next to it. This PR neither widens nor narrows the window.
3. **Under-warning on a failed opt-in.** This is the issue's explicit intent, not a regression. Worth naming the sharpest version of it: with stale page data (user is listed server-side but the page rendered `optedIn: false`), a failed opt-in now says "you have not been added" when they may in fact be listed. But the switch itself already asserted OFF in exactly that scenario on `main` — the prose now *agrees* with the control instead of contradicting it. No new privacy surface; a message and a switch that disagree were the worse failure.
4. **The `role="alert"` gap the implementor flagged.** It mattered enough to be worth closing, and it is now closed — the strings were read out of the rendered alert node, not off the handler.

### Non-blocking observations

- **Behaviour change nobody flagged:** `{ error: "" }`. On `main`, `setError("")` is falsy, so `{error && ...}` rendered **nothing** — a silent failure with a reverted switch and no explanation. On this branch it renders `. You have not been added to the leaderboard.` with an orphan leading period. Unreachable from the current action, and honestly the better of the two behaviours (an ugly sentence beats silence on a privacy control), but it is a change from `main` that the "wording only" framing does not cover. No action needed.
- `next lint` prints a Next.js 16 deprecation warning. Pre-existing, repo-wide, unrelated.

### What remains unexercised

Stating this plainly because approval here means it merges:

- **No browser.** jsdom + real React + real MUI is much closer than a stubbed harness, but it is not Chrome. I did not load `/games/wordleverse/leaderboard` with a signed-in session against a live `DATABASE_URL`.
- **Nothing server-side ran.** The action is stubbed, so `getCurrentUser()`, the Prisma write, `revalidatePath` and the real `router.refresh()` re-render were never exercised. This PR does not touch any of them.
- **`npm run build` not run** — it needs a reachable `DATABASE_URL` for `prisma generate`. `npm run lint` is the repo's stated gate and it passes.
- **Nothing guards this going forward.** The repo has no test runner and the plan deliberately ruled out adding one; both my harness and the implementor's are throwaway. A future edit to that template has nothing to catch it. That is a standing property of this repo, not a defect in this PR.

Issue #144, opened 2026-08-15
feat(RC): online multiplayer and site integration


csalinas-dev:
**Base branch: `epic/race-condition`** (epic #141). Open the PR against that branch, not `main`.

Part 3 of #141 (epic). Depends on the board issue (part 2) — do not start until it
is merged.

Put **Race Condition** in a room, using the shared realtime core exactly as Edge
Case and Connect 404 do: a pure `{ id, maxPlayers, colors, createState, reducer }`
definition registered in `src/lib/realtime/registry.js`. Read
`src/lib/realtime/README.md` first — it is the contract.

Hard constraints, all of which the core already establishes:
- **No game rules under `src/lib/realtime/`.** The definition calls into the
  game's `_lib/` rather than restating any rule; one implementation,
  server-authoritative.
- Name the definition file `multiplayer.js`, not `game.js` — a `Game.jsx` beside
  it collides on case-insensitive filesystems.
- MySQL JSON columns do not preserve object key order, so nothing in the room's
  state may depend on it.
- Identity is the browser's player token; request bodies never name a slot, and
  the token never appears in a URL.
- The hotseat mode from part 2 must keep working with zero network calls in that
  branch, filling the same Context.

Also finish the site integration: the game card and thumbnail on `/games`, and
whatever `/privacy` or terms copy the other multiplayer games carry (see #107) if
this game adds anything they do not already cover.

### Done when
- Two browsers can play a full game in one room, including rematch, and a third
  arrival is handled the way the other games handle it.
- Reconnecting after a refresh returns the player to their seat.
- Leaving and disconnecting are distinguished, using `src/lib/realtime/absence.js`
  rather than testing the flags directly.
- The game appears on `/games` with artwork consistent with the existing cards.

---

## 🤖 Implementation Plan

_Written by the architect agent on 2026-08-15 23:03 UTC. The implementor builds exactly this._

### Goal

Race Condition becomes playable online: two browsers, one four-character room code, a
full game with a rematch, a third arrival that spectates, a refresh that returns you to
your seat, and a distinction on screen between somebody who left and somebody whose
phone locked. The game also appears on `/games` and in the nav, and the privacy/terms
copy stops naming only two of the site's four online games. The hotseat game from #143
keeps working with zero network calls in that branch.

**Base branch is `epic/race-condition`, not `main`.** Branch from it and open the PR
against it. `main` does not have the engine or the board.

---

### Approach

[...]

Issue #143, opened 2026-08-15
feat(RC): animated Orbito board and hotseat play


csalinas-dev:
**Base branch: `epic/race-condition`** (epic #141). Open the PR against that branch, not `main`.

Part 2 of #141 (epic). Depends on #142 — do not start until #142 is merged; this
issue consumes the engine's API.

Build the playable **Race Condition** board at its route under
`src/app/(pages)/games/`, in same-device hotseat mode only. Network code comes in
part 3.

The look should sit alongside Connect 404 — same visual language, same component
conventions, same Context shape (`{ state, dispatch, canPlay, canReset, online }`)
so part 3 can fill that context from a room without rewriting a single cell,
status, or toolbar component. Read how Tic Tac Overflow and Connect 404 do this
before designing anything new.

The animation is the point of this game: the orbit rotation at the end of each turn
must be legible as motion, so a player can see where every marble went rather than
having the board teleport. The optional move step and the placement step need to
read as two distinct beats of one turn.

All rules come from #142's `_lib/`. If the board needs a rule the engine does not
expose, extend the engine — never re-implement it in a component.

### Done when
- The route is reachable, playable end to end by two people on one device, and
  reports wins and ties correctly.
- The orbit rotation is animated and readable; no state teleports.
- Every rule decision traces to a call into `_lib/`.
- The board renders correctly at 440px and 1280px, and does not introduce viewport
  units in a block axis (see #138 for why).

---

## 🤖 Implementation Plan

_Written by the architect agent on 2026-08-15 21:11 UTC. The implementor builds exactly this._

### Goal

`/games/race-condition` becomes a playable, animated Orbito board: two people on one
device take whole turns (optional slide of an *opponent's* marble → place your own →
the board orbits), and the orbit is watched rather than jumped over. Every rule
decision is a call into `src/app/(pages)/games/race-condition/_lib/`. The context and
file layout are the ones #144 will fill from a room without rewriting a component.

### Approach

**The board is a grid of 16 socket/button cells with a separate, absolutely
positioned marble layer on top.** Marbles get synthetic stable ids, so React keeps
the same DOM node across an orbit and a `transition: transform` slides it. This is
the whole animation design, and it works because of a fact I verified rather than
assumed:

> **Every step of both orbit tracks is exactly one orthogonally adjacent cell.**
[...]

Issue #141, opened 2026-08-15
feat(race condition): build game


csalinas-dev:
**Epic branch: `epic/race-condition`** — every sub-issue PR (#142, #143, #144) targets this branch, not `main`. When all three have merged, a final PR merges `epic/race-condition` into `main`.

---

https://flexiqgames.com/en/product/orbito/

Same multiplayer room as the other games, like Connect 404.

I want a similar look as Connect 404, but obviously the board should be an Orbito board. Same game mechanics as Orbito.

Split into sub-issues, as needed.

csalinas-dev:
Epic branch created: `epic/race-condition`, cut from `main` at 0b02794.

Correction to how this epic has run so far: #142 (the engine) was merged straight into `main` before this rule existed, so the engine is already on the default branch and is *not* part of the epic branch's eventual diff. `epic/race-condition` was cut from `main` after that merge, so it contains the engine — nothing is lost, but the final integration PR will show only the board (#143) and online (#144) work. Flagging it rather than quietly letting the epic PR look smaller than the feature.

csalinas-dev:
All three sub-issues merged. Integration PR #148 (epic/race-condition -> main) is up for review.

csalinas-dev:
Changes requested on https://github.com/csalinas-dev/csalinas-dev-site/pull/148: rc-ending.mjs's generated page script never reaches the board on the merged branch — #144's mode picker sits in front of it and the script bails with exit code 0. Feature itself plays correctly hotseat and online; no regression in the other three games.

csalinas-dev:
Revisions pushed (836f907), ready for re-review. rc-ending.mjs's page script now drives the mode picker into the hotseat board and every return carries `ok`. All three endings re-verified in my own tree, not assumed from the review: orbited-in 12 turns / ring [0,5,10,15], simultaneous 15 / both columns ringed, exhausted 16 / 16 marbles, ok:true on all three. Negative control returns ok:false. No game code touched, so the hotseat/online/blast-radius verification stands. shot.mjs's exit code left to #150.

csalinas-dev:
[...]

PR #145, opened 2026-08-15
feat(RC): Orbito engine — orbit rotation, moves, and win detection

### LINKED ISSUE #142: feat(RC): Orbito engine — orbit rotation, moves, and win detection
> **Correction (architect, verified against the publisher rulebook):** the move step below is wrong. You move one of your OPPONENT'S marbles, not "any single marble". The plan in this issue is authoritative.

Part 1 of #141 (epic). Build the pure rules engine for **Race Condition**, the
site's Orbito variant, with no UI and no network code.

Orbito is a two-player 4x4 abstract. Official rules and reference implementation
to check against: https://flexiqgames.com/en/product/orbito/ — **verify the exact
mechanics against the publisher's rules before writing the plan**; do not encode
them from memory. The parts that matter and are easy to get wrong: what a turn
consists of (an optional move of any single marble to an adjacent empty square,
then placing one of your own), the orbit rotation that happens at the *end* of
every turn along a fixed track, when win detection runs relative to that rotation,
and what happens when both players end up with four in a row simultaneously.

Follow the pattern Edge Case set and Connect 404 followed: **all** rules live in a
pure, synchronous, dependency-free `_lib/` module under the game's route directory,
so the server and every client run the same code unchanged. No rule may be restated
in a component or in the realtime core. Document the board/orbit-track coordinate
convention at the top of the module the way `edge-case/_lib/edges.js` does, so
nobody re-derives it later.

Ties must be reported as ties, never broken.

### Done when
- A `_lib/` module exports state creation, the legal-move check, the turn
  application (move → place → orbit), and result/win detection, all pure.
- The orbit track is defined once, derived or documented, and covered by tests or a
  verification script that shows a marble returning to its start after a full cycle.
- Simultaneous four-in-a-row is handled explicitly, per the published rules.
- No React, no Prisma, no imports from `src/lib/realtime/` in the engine.

---

## 🤖 Implementation Plan

_Written by the architect agent on 2026-08-15 20:26 UTC. The implementor builds exactly this._

### Goal

`src/app/(pages)/games/race-condition/_lib/` exists and is the single, pure,
dependency-free authority for Race Condition (the site's Orbito). It creates
state, validates and applies a whole turn (move → place → orbit), detects wins
only after the orbit, reports ties as ties, and exposes a reducer the shared
realtime core can call server-side unchanged. No UI, no route, no network.

---

## The published rules — verified, not remembered

I extracted the text of the publisher's rulebook
(<https://flexiqgames.com/wp-content/uploads/2023/04/230308-FlexiQ-FXG-502-Orbito-Game-Rules-A4-UK-HR-v1.pdf>,
linked from <https://flexiqgames.com/en/product/orbito/>). **Build to these
sentences, not to the summary in the issue body** — the issue body is wrong on
one point, flagged below. Quoted verbatim:

> CONTENTS: 1 shifting 4x4-game board with Orbito-button, 8 white marbles,
> 8 black marbles

> Each player in turn will add one of their marbles to the game board. Each turn
> consists of 3 steps:

> **1 — MOVE ONE OF YOUR OPPONENT'S MARBLES TO AN ADJACENT, EMPTY SQUARE.**
> • This step is not possible during the first turn of the game (as there is no
> opponent's marble on the board yet) or if all positions adjacent to your
> opponent's marbles are occupied. • This step is optional. You can choose not
> to move an opponent's marble. • The marble can only be moved to an adjacent
> square and only horizontally or vertically, not diagonally.

> **2 — PLACE ONE OF YOUR OWN MARBLES ONTO ANY FREE POSITION ON THE BOARD.**
> Once placed, you cannot move your own marbles anymore. Only your opponent can!

> **3 — TO FINISH YOUR TURN, YOU MUST PRESS THE ORBITO-BUTTON!** This will cause
> **ALL** marbles to shift 1 position on the board! **Both the inner & outer
> orbits rotate counter-clockwise.**

> Only after pressing the Orbito-button, a valid 4-in-a-row can be created. So
> never during a player's turn!

> The winner is the first player to have 4 marbles of their colour aligned
> horizontally, vertically or diagonally! It is possible that after pressing the
> Orbito-button, a player creates a sequence of 4 in their **opponent's** colour
> and makes their opponent win!

> If, after pressing the Orbito-button, both players simultaneously have a
> sequence of 4, it's also a draw!

> No player has a sequence of 4 after the last marble is placed onto the board
> and the Orbito-button has been pressed! In this case, the Orbito-button needs
> to be pressed 5 more times. The first player to get a sequence wins the game.
> If there is no sequence after 5 times, it's a draw!

**⚠️ Correction to the issue body.** The issue says a turn begins with "an
optional move of any single marble". It does not. **You may only move one of
your OPPONENT'S marbles**, and only orthogonally, and only onto an empty square.
Moving your own marble is illegal and must be refused with its own error code.

### Ambiguities in the published rules, and the interpretation I chose

The rulebook does not settle these. I picked one each; do not silently pick a
different one.

1. **Does the simultaneous-lines draw apply during the 5 overtime presses?**
   The rulebook states the simultaneous rule for "after pressing the
   Orbito-button" generally, and the overtime presses are Orbito-button presses.
   **Chosen: yes** — a press that completes lines for both colours is a draw
   (`drawReason: "simultaneous"`) whether it is the turn's first press or its
   sixth.
2. **Who "presses" the 5 overtime times?** The rulebook says only that the
   button "needs to be pressed 5 more times"; nobody chooses anything and no
   marble moves, so the outcome is fully determined the moment the last marble
   is placed. **Chosen: the engine resolves all of it inside the same
   `playTurn` call** that fills the board — that turn's `spins` runs 1…6 and the
   game is over when `playTurn` returns. *Rejected:* a separate `SPIN` action.
   It would invent a turn-owner the rules never name, and online it would let a
   disconnect freeze a game whose result is already decided.
3. **"The first player to get a sequence wins" during overtime** — this means
   the first press at which some colour completes a line, not an alternation of
   pressers. **Chosen: stop at the first press that completes any line**; if
   exactly one colour has lines that player wins, if both do it is a draw.
4. **Board orientation of "counter-clockwise".** Fully determined for a square
   grid viewed from above; the screen view is the same view. Verified table
   below — no ambiguity, but do not re-derive it.

---

### Approach

One engine module, laid out exactly like the two that already exist
(`edge-case/_lib/` and `connect-404/_lib/`): pure, synchronous, zero imports
outside itself, a rejected move returns the state **by reference** plus an
`{ code, message }`, a bad *construction* throws.

**A turn is one atomic action**, `{ type: "PLAY TURN", move, place }`, applied by
`playTurn(state, { move, place }, slot)` which runs move → place → orbit →
win-scan in a single call. *Rejected:* separate `MOVE MARBLE` / `PLACE MARBLE`
actions with a sub-phase on the state. The atomic form is what structurally
guarantees the rule "a 4-in-a-row is only valid after the button" — there is no
representable state in which a placement has landed and the orbit has not — and
it costs the realtime core one revision per turn instead of two, with no
half-finished turn to strand on the server when a player disconnects mid-turn.
The cost is that #143 stages the move/placement client-side before committing;
that is why `legalMoves`/`canMove` are exported.

**Overtime resolves inside that same call** (ambiguity 2 above), so the state
needs no phase field. `lastTurn.spins` tells #143 how many rotations to animate.

---

- **Read these two first and copy their shape, tone and JSDoc density.** They are
  the pattern, not merely similar code:
  - `src/app/(pages)/games/connect-404/_lib/` — closest sibling: two players,
    fixed board, `createState({ slots, first })` / `resetState`, `MoveError`
    frozen enum, `reject(state, code)` returning the same reference, `slotOf`
    in `reducer.js` accepting either a bare slot or the core's player object.
  - `src/app/(pages)/games/edge-case/_lib/edges.js` — the box-drawing header
    comment that documents a coordinate convention. Yours must be as explicit.
  - `src/app/(pages)/games/edge-case/_lib/getResult.js` — how a tie is reported
    rather than broken.
- **`state.slots` must be an ARRAY in join order**, and nothing may derive turn
  order from `Object.keys` of anything. Room state round-trips through a MySQL
  JSON column and MySQL does not preserve object key order — see the comment in
  `connect-404/_lib/createState.js`. Same reason `getResult` returns `hands` as
  an array, not an object keyed by slot.
- **Who consumes this:** #143 (board + hotseat) and #144 (`multiplayer.js` +
  `src/lib/realtime/registry.js`). See
  `src/app/(pages)/games/connect-404/multiplayer.js` for how a definition wraps
  an engine — it *calls* the engine's reducer and never restates a rule. Do not
  write that file in this PR; just make sure `reducer(state, action, player)`
  can be called by it.
- **There is no test runner in this repo.** `package.json` has only
  `dev`/`build`/`start`/`lint`. Verification is a committed Node script — see
  the harness section below; I have already proved the mechanism works.
- Directory `src/app/(pages)/games/race-condition/` does not exist yet. `_lib`
  is a Next.js private folder, so adding it creates no route.

---

### The board convention (verified — do not re-derive)

```
INDEXING — a 4x4 board stored ROW-MAJOR as a flat 16-cell array.

    index = row * 4 + col       row 0 is the TOP, col 0 is the LEFT
    row grows DOWNWARD, col grows RIGHTWARD

          col  0    1    2    3
        row 0  0    1    2    3
        row 1  4    5    6    7
        row 2  8    9   10   11
        row 3 12   13   14   15

`state.cells[i]` is the slot whose marble sits there, or null. There is no
second representation of the board; there are no piece objects.

ADJACENCY for step 1 is ORTHOGONAL ONLY — index ±4 (with row bounds) and
index ±1 (only when both cells share a row). Never diagonal.
```

### The orbit convention (verified numerically — do not re-derive)

```
TWO ORBITS, both turning COUNTER-CLOCKWISE by exactly one position per press.
Empty squares travel too: an orbit is a permutation of all 16 cells, not a
move of the occupied ones.

  OUTER — the 12 border cells, in travel order (each cell's marble moves to
  the NEXT entry, and 1 wraps to 0):

      [0, 4, 8, 12, 13, 14, 15, 11, 7, 3, 2, 1]

      down the left column, right along the bottom, up the right column,
      left along the top — which is what counter-clockwise means with row 0
      on top.

  INNER — the 4 centre cells, same direction:

      [5, 9, 10, 6]

One press, drawn:

      A B C D          B C D H
      E F G H    ->    A G K L
      I J K L          E F J P
      M N O P          I M N O

Checked: the two tracks are disjoint and cover all 16 cells; every outer cell
returns to its start after exactly 12 presses, every inner cell after 4, and
the whole board is identical to its start after 12 presses and not before.
```

**Derive both tracks in code from the grid walk** (loop the left column top to
bottom, the bottom row left to right, the right column bottom to top, the top
row right to left) rather than pasting the literals — then assert in the verify
script that they equal the arrays above. That way the numbers are checked, not
trusted.

### The 10 winning lines (verified)

Exactly 4 rows + 4 columns + 2 diagonals, each exactly 4 cells — a run of five
is impossible here, so unlike Connect 404 there is no partial-run logic:

```
[0,1,2,3] [4,5,6,7] [8,9,10,11] [12,13,14,15]
[0,4,8,12] [1,5,9,13] [2,6,10,14] [3,7,11,15]
[0,5,10,15] [3,6,9,12]
```

Derive these in code too, and assert against this list.

---

### The state shape (load-bearing — #143 and #144 build on it)

```js
{
  size: 4,                     // SIZE; cells = size * size
  cells: (slot|null)[16],      // row-major, index = row * 4 + col, row 0 TOP
  slots: [slotA, slotB],       // join order, ARRAY (MySQL key-order rule)
  turn: slot,                  // whose turn; on a finished board, whoever ended it
  lastTurn: {                  // null before the first turn
    by: slot,
    move: { from: number, to: number } | null,   // step 1, or null if declined
    place: number,                                // step 2, cell index
    cellsBeforeOrbit: (slot|null)[16],            // board after step 2, before any press
    spins: number,                                // presses applied: 1 normally, up to 6
                                                  // on the turn that fills the board
  } | null,
  winner: slot | null,
  winningLines: number[][] | null,  // every completed line at the deciding press.
                                    // On a simultaneous draw this holds BOTH
                                    // colours' lines; colour each by cells[line[0]].
  draw: boolean,
  drawReason: "simultaneous" | "exhausted" | null,
  finished: boolean,                // winner !== null || draw
}
```

Notes the implementor must honour:

- **Marbles in hand are DERIVED, not stored**: `marblesInHand(state, slot)` is
  `MARBLES_PER_PLAYER - marblesOnBoard(state, slot)`. Marbles never leave the
  board, so this is always exact and cannot drift. Do not add a `reserves` field.
- **`cellsBeforeOrbit` + `spins` is the whole animation contract for #143**: the
  board renders the pre-orbit position, then calls the exported `orbit()`
  `spins` times to walk forward one press at a time. That is why `orbit` is
  exported and why `cellsBeforeOrbit` is stored rather than left to be
  back-derived.
- `spins > 1` happens only on the turn that fills the 16th square — that is how
  a renderer knows it is showing overtime.

---

### Changes

All paths under `src/app/(pages)/games/race-condition/`.

1. **`_lib/constants.js`** — `SIZE = 4`, `CELLS = 16`, `LINE_LENGTH = 4`,
   `MARBLES_PER_PLAYER = 8`, `MIN_PLAYERS = 2`, `MAX_PLAYERS = 2`,
   `OVERTIME_SPINS = 5`, `PLAY_TURN = "PLAY TURN"`, and a frozen `MoveError`:
   `Finished`, `UnknownSlot`, `NotYourTurn`, `BadTurn` (malformed payload),
   `OutOfRange`, `Occupied` (target square already holds a marble — used for
   both the move destination and the placement), `EmptySource` (nothing on
   `move.from`), `OwnMarble` (step 1 may only move an opponent's marble),
   `NotAdjacent`, `UnknownAction`. Each with the one sentence a player reads.
   Comment `MARBLES_PER_PLAYER * 2 === CELLS` — that identity is why the board
   fills exactly and why there is no "out of marbles" state to reach.

2. **`_lib/board.js`** — the INDEXING header comment above, verbatim in the
   `edges.js` style, then: `cellIndex(row, col)`, `cellCoords(index)`,
   `isCellInRange(index)`, `cellAt(board, index)`, `isEmpty(board, index)`,
   `emptyCells(board)`, `isBoardFull(board)`, `marblesOnBoard(board, slot)`,
   `marblesInHand(state, slot)`, `neighbors(index)`, `areAdjacent(a, b)`.
   `areAdjacent` must reject the row-wrap case: index 3 and index 4 are `±1`
   apart but are not neighbours.

3. **`_lib/orbit.js`** — the ORBIT header comment above, verbatim. Derive
   `OUTER_TRACK` and `INNER_TRACK` from the grid walk, build a single flat
   permutation from them, and export `orbit(cells)` (one press) and
   `unorbit(cells)` (its inverse — the renderer and the verify script both want
   it). `orbit` returns a new array and never mutates.

4. **`_lib/lines.js`** — derive `LINES` (10 × 4 indices), export
   `completedLines(cells)` → every line whose four cells hold the same non-null
   slot, and `linesFor(cells, slot)`. Keep it a whole-board scan: after an orbit
   a line can appear anywhere, so the Connect 404 trick of scanning only from
   the cell just played is **wrong here**.

5. **`_lib/createState.js`** — `createState({ slots, first })`, throwing (like
   the siblings) on a non-array, wrong count, duplicates, or a `first` not in
   `slots`. `resetState({ slots, winner })` → winner opens the next game, join
   order after a draw.

6. **`_lib/moves.js`** — `canMove(state, from, to, slot)` returning a
   `MoveError` code or `null`, and `legalMoves(state, slot)` →
   `[{ from, to }, …]` (what #143 highlights). Rules: `from` holds a marble,
   that marble is **not** `slot`'s, `to` is in range, `to` is empty, `to` is
   orthogonally adjacent to `from`. On turn one `legalMoves` returns `[]` for
   free because no opponent marble is on the board.

7. **`_lib/playTurn.js`** — `reject(state, code)` (returns the **same** state
   reference) and `playTurn(state, { move, place }, slot)`:
   - refuse: finished, unknown slot, not your turn, malformed payload;
   - apply step 1 if `move` is given, validating via `canMove`;
   - apply step 2 onto the post-move board — `place` must be in range and empty
     *after* the move, so placing onto the square the opponent's marble just
     vacated is legal and placing onto `move.to` is not;
   - snapshot that board as `cellsBeforeOrbit`;
   - press: apply `orbit`, `spins = 1`, scan `completedLines`. If lines exist,
     settle (one colour → that slot wins even if it is the opponent; two colours
     → `draw` with `drawReason: "simultaneous"`);
   - if no lines **and** the board is now full, press up to `OVERTIME_SPINS = 5`
     more times, settling at the first press that completes any line; if none
     do, `draw` with `drawReason: "exhausted"` and `spins === 6`;
   - otherwise the turn passes to the other slot. On a finished board `turn`
     stays with the player who ended it, matching Connect 404.

8. **`_lib/getResult.js`** — reads the decided fields back off the state, does
   **not** re-decide the game, and adds what a status line wants:
   `{ winner, draw, drawReason, winningLines, finished, turn, placed, remaining,
   spins, hands: [{ slot, onBoard, inHand }, …] }`. `hands` is an **array** in
   `slots` order. Document that `draw` is read before `winner`, and that a tie
   is never broken.

9. **`_lib/reducer.js`** — `slotOf(player)` accepting either a bare slot or the
   core's player object (copy `connect-404/_lib/reducer.js`), switching on
   `PLAY_TURN` → `playTurn`, defaulting to `reject(state, MoveError.UnknownAction)`.
   Named and default export.

10. **`_lib/index.js`** — barrel re-exporting everything above, headed by the
    module doc comment in the Connect 404 style: the tagline, "single source of
    truth / pure / synchronous / dependency-free", the state shape block above,
    a pointer to `board.js` and `orbit.js` for the conventions ("read it there;
    do not re-derive it"), the by-reference-rejection contract, and the
    correction that **only an opponent's marble may be moved**. Suggested
    tagline for the header, matching the siblings' cadence:
    *"Everything moves at once. Line up first."* (#143 owns the displayed
    tagline and may differ; the header is documentation.)

11. **`.agent/scripts/lib/esm-resolver.mjs`** (new) — a ~20-line Node
    `module.register` resolve hook that retries a failed **relative** specifier
    with `.js`, `.jsx`, `/index.js`, `/index.jsx`. This is what lets bare Node
    import a `_lib` that uses the repo's extensionless imports. **I have
    verified this works**: with the hook registered, Node 24 imports
    `connect-404/_lib/index.js` and runs `createState` unchanged. Keep it
    generic so the next engine reuses it.

12. **`.agent/scripts/verify-race-condition.mjs`** (new) — the verification
    harness, run with plain `node`, exiting non-zero on any failure and printing
    a one-line-per-check summary. Cases, at minimum:
    - **Orbit geometry.** Tracks equal the verified literals; disjoint; cover
      all 16. `orbit` moves 0→4, 3→2, 12→13, 15→11, 5→9, 6→5 (an explicit
      table, so a future refactor that flips the direction fails loudly).
      `unorbit(orbit(b))` is `b`. **Every cell returns to its start after 12
      presses and not before; inner cells return after 4.** (This is the "full
      cycle" check the issue asks for.)
    - **Lines.** Exactly 10, each of length 4, equal to the verified list.
    - **An independently written win scanner** (write a second one inside the
      script from scratch — do not import `completedLines`) agrees with the
      engine on the completed lines of several thousand random boards.
    - **A win is never reported before the orbit.** Construct a position where
      the placement itself makes 4-in-a-row but the following orbit does not,
      and assert `finished === false` and `winner === null`.
    - **You can lose by your own hand.** Construct a turn whose orbit completes
      a line for the *opponent* only, and assert `winner` is the opponent, not
      the mover.
    - **Simultaneous lines are a draw.** `winner === null`, `draw === true`,
      `drawReason === "simultaneous"`, and `winningLines` holds both colours'
      lines.
    - **Overtime.** A board filled with no line ends the game inside the last
      `playTurn`: `finished === true`, `lastTurn.spins === 6`,
      `drawReason === "exhausted"` — or a win found at press *k* leaves
      `spins === k` and stops there.
    - **Legality.** Refused, each with its own code: moving your own marble; a
      diagonal move; a move to an occupied square; a move across the 3→4 row
      wrap; any move on turn one; placing onto an occupied square; placing onto
      `move.to`; playing out of turn; playing after the game is finished.
      **Allowed:** placing onto the square `move.from` just vacated.
    - **Rejection is by reference**: `result.state === state` on every refusal.
    - **Random self-play**, ≥2000 games: never more than 8 marbles of a colour
      on the board, never more than 16 placements, `spins` never exceeds 6,
      every terminal state has exactly one of `winner`/`draw` set, and
      `finished` is never true with neither.

13. **`CLAUDE.md`** — add a bullet in the existing games section: where the
    engine lives, that all rules live in `_lib/` and none may be restated in a
    component or the realtime core, the correction that only an **opponent's**
    marble may be moved, that both orbits turn counter-clockwise and the
    conventions are documented in `board.js`/`orbit.js`, that a turn is one
    atomic `PLAY TURN` action with overtime resolved inside it, and that
    `node .agent/scripts/verify-race-condition.mjs` is how the engine is checked.

---

### Out of scope

- **No route, no `page.js`, no components, no artwork, no `games.js` entry, no
  nav.** #143 owns all of that.
- **No `multiplayer.js` and no edit to `src/lib/realtime/registry.js`.** #144
  owns that. Nothing under `src/lib/realtime/` changes in this PR.
- No React, no Prisma, no `next/*`, no lodash, no import from `@/` — the engine
  imports nothing but its own siblings.
- No AI/bot opponent, no move history beyond `lastTurn`, no board sizes other
  than 4x4, no player counts other than 2.
- Do not add a test framework. The verify script is the deliberate choice
  because this repo has no runner, and adding one is its own decision.

---

### Acceptance criteria

- [ ] `src/app/(pages)/games/race-condition/_lib/` exports `createState`,
      `resetState`, `playTurn`, `reducer`, `getResult`, `legalMoves`, `canMove`,
      `orbit`, `unorbit`, `completedLines`, the board helpers, and the constants
      including `PLAY_TURN` and `MoveError`.
- [ ] The engine imports nothing outside `_lib/` — no React, no Prisma, no
      `src/lib/realtime/`, no `@/` alias, no third-party package.
- [ ] `board.js` and `orbit.js` each open with an explicit convention comment in
      the style of `edge-case/_lib/edges.js`; `index.js` documents the state
      shape and points at them.
- [ ] Only an **opponent's** marble can be moved in step 1, orthogonally, onto
      an empty square; moving your own marble is refused with its own code.
- [ ] A 4-in-a-row is only ever a win **after** the orbit — a placement that
      lines up four and is then broken by the orbit does not end the game.
- [ ] A player can complete a line in the opponent's colour and the opponent wins.
- [ ] Both colours lining up on the same press is reported as a draw, never
      broken; `winningLines` shows both.
- [ ] Filling the board with no line runs the 5 extra presses inside the same
      `playTurn`, stopping at the first press that completes a line, otherwise
      ending as a draw with `drawReason: "exhausted"` and `lastTurn.spins === 6`.
- [ ] A refused move returns the **same state reference** plus
      `{ code, message }`, and never throws; `createState` throws on a bad roster.
- [ ] `node .agent/scripts/verify-race-condition.mjs` passes and its output is
      quoted in the PR body, including the full-orbit-cycle result and the
      independent-scanner agreement count.
- [ ] `npm run lint` is clean.
- [ ] `CLAUDE.md` documents the engine.

---

### Verification

```bash
npm run lint
```

Both verified to work in this repo: `lint` is a real script in `package.json`,
and I confirmed the `module.register` resolve-hook mechanism actually imports a
`_lib` engine under bare Node 24 (`createState` from `connect-404/_lib/index.js`
ran unchanged). There is no test runner; do not invent an `npm test`.

`npm run build` will not exercise the engine at all — nothing imports `_lib`
until #143 — so it is not a meaningful check here.

---

### Risks

- **Getting the orbit direction backwards** is the highest-value bug and the
  hardest to see: a mirrored board still plays a legal-looking game. The
  explicit 0→4 / 3→2 / 12→13 / 15→11 / 5→9 / 6→5 assertion table in the verify
  script is the guard. Do not weaken it to "the board changed".
- **Scanning for a win from only the placed cell** (the Connect 404 shortcut)
  silently misses most wins here, because the orbit moves everything. The
  independent-scanner check catches it.
- **Checking for a win before the orbit** would make the game end a press early
  and would contradict the rulebook's most emphatic sentence. There is a
  dedicated check for it.
- **Row-wrap adjacency** (3↔4, 7↔8, 11↔12) is the classic flat-array bug and
  would let a marble teleport across the board in step 1.
- **Object key order.** Anything keyed by slot that gets *iterated* will scramble
  once #144 stores this state in MySQL JSON. `slots` and `hands` are arrays for
  exactly this reason; keep it that way.
- **Overtime running away.** Cap the loop at `OVERTIME_SPINS`; the full board's
  orbit has period 12, so it genuinely can cycle without ever producing a line.

csalinas-dev:
Closes #142

Part 1 of #141. The pure rules engine for **Race Condition** (the site's Orbito) — no UI, no route, no network. Follows the pattern Edge Case set and Connect 404 followed: all rules in a pure, synchronous, dependency-free `_lib/` under the game's route directory, so the server and every client run the same code unchanged.

## What changed

Read in this order.

- **`_lib/constants.js`** — board/marble/overtime numbers, the frozen `MoveError` enum, `PLAY_TURN`, and a frozen `DrawReason`. `MARBLES_PER_PLAYER * 2 === CELLS` is commented as the identity the endgame rests on: the board fills exactly as the last marble lands, so there is no "out of marbles" state.
- **`_lib/board.js`** — the flat row-major indexing convention as a header comment in the `edge-case/_lib/edges.js` style, then the geometry helpers. `areAdjacent` compares rows and columns rather than flat indices, so the 3↔4 / 7↔8 / 11↔12 row wrap is not a move.
- **`_lib/orbit.js`** — the orbit convention as a header comment, then `OUTER_TRACK`/`INNER_TRACK` **derived** by a counter-clockwise ring walk (one `ringTrack(inset)` produces both), flattened into a single 16-entry permutation, plus `orbit`/`unorbit`. The plan's literals live in the verify script as the assertion, not here as the implementation.
- **`_lib/lines.js`** — the ten lines derived from the grid; `completedLines` is a **whole-board** scan, with a comment saying why the Connect 404 "scan from the cell just played" shortcut is wrong here.
- **`_lib/createState.js`** — `createState({ slots, first })` / `resetState`, throwing on a bad roster like its siblings. `slots` is an array, for the MySQL-key-order reason.
- **`_lib/moves.js`** — `canMove` / `legalMoves`, quoting the rulebook on step 1.
- **`_lib/playTurn.js`** — the whole turn as one call: slide → place → press → scan, with overtime resolved inside the same call.
- **`_lib/getResult.js`** — reads the decided fields back, adds counts and `hands` (an **array** in `slots` order). Does not re-decide the game and never breaks a tie.
- **`_lib/reducer.js`** — `PLAY_TURN` → `playTurn`, `slotOf` accepting either a bare slot or the core's player object. Not registered anywhere; #144 owns that.
- **`_lib/index.js`** — barrel plus the module doc comment: state shape, pointers to the two convention headers, the by-reference rejection contract, and the correction below.
- **`.agent/scripts/lib/esm-resolver.mjs`** — a generic Node resolve hook that retries failed *relative* specifiers with `.js`/`.jsx`/`/index.js`/`/index.jsx`, so bare `node` can import this repo's extensionless imports. Bare specifiers are left alone so a missing dependency still fails loudly.
- **`.agent/scripts/verify-race-condition.mjs`** — the checks. This repo has no test runner.
- **`CLAUDE.md`** — five bullets covering the engine, the opponent's-marble rule, the orbit conventions, the atomic turn + overtime, and how it is verified.

### The rule the issue's summary got wrong

The issue body says a turn begins with "an optional move of any single marble". The architect corrected this against the publisher's rulebook and I built the plan, not the summary: **you may only move one of your OPPONENT'S marbles** ("Once placed, you cannot move your own marbles anymore. Only your opponent can!"), orthogonally, onto an empty square. Reaching for your own is refused with its own code, `MoveError.OwnMarble`.

### Two design points worth a second look

**A turn is one atomic action.** `playTurn(state, { move, place }, slot)` runs all three steps. This is what structurally guarantees the rulebook's most emphatic sentence — "Only after pressing the Orbito-button, a valid 4-in-a-row can be created" — because no state exists in which a marble has landed and the button has not. It also costs the realtime core one revision per turn instead of two.

**Overtime resolves inside that same call.** Nobody chooses anything during the five extra presses and no marble moves, so the outcome is decided the instant the sixteenth marble lands. A separate `SPIN` action would have to invent a turn-owner the rules never name, and online it would let one disconnect freeze a game whose result was already determined.

Both were the plan's stated choices with its stated rejected alternatives; I have not re-litigated them.

## How it was verified

    $ node .agent/scripts/verify-race-condition.mjs ; echo "exit $?"

    Constants — pinned to the rulebook, not to themselves
      ok    the board, the marbles and the overtime presses are the published ones — 4x4, 8 marbles each, 5 overtime presses

    Orbit — tracks, direction and cycle
      ok    outer track is the verified counter-clockwise border walk
      ok    inner track is the verified counter-clockwise centre walk
      ok    the two tracks are disjoint and cover all 16 squares
      ok    a press moves marbles the counter-clockwise way (explicit table) — 6 of 16 destinations pinned
      ok    one press, drawn — the whole 16-square permutation
      ok    orbit does not mutate its input
      ok    unorbit is the exact inverse of orbit — 500 random boards round-tripped
      ok    every outer square returns home after exactly 12 presses — 12 squares, period 12 each
      ok    every inner square returns home after exactly 4 presses — 4 squares, period 4 each
      ok    the whole board repeats after 12 presses and not before — full cycle = 12 presses

    Lines — the ten ways to win
      ok    there are exactly 10 lines, each 4 squares long
      ok    the derived lines are the verified 4 rows + 4 columns + 2 diagonals
      ok    the win scan agrees with an independently written one — 5000 random boards agree, 1850 of them holding at least one line

    Rules — the press decides, not the placement
      ok    a line made by the placement and broken by the press does NOT win — 4-in-a-row before the press, none after, game continues
      ok    you can hand the win to your opponent — 0 pressed the button, 1 won
      ok    both colours lining up on one press is a draw, and it is not broken — both lines reported, drawReason "simultaneous"

    Overtime — the five extra presses
      ok    a full board with nothing found runs all 6 presses and is a draw — 6 presses, nothing found
      ok    overtime stops at the first press that completes a line — stopped on press 3 of a possible 6
      ok    spins is 1 on an ordinary turn

    Legality — step 1 moves your OPPONENT'S marble
      ok    moving one of your OWN marbles — → own-marble
      ok    moving diagonally — → not-adjacent
      ok    moving onto an occupied square — → occupied
      ok    moving across the flat-array row wrap (3 → 4) — → not-adjacent
      ok    moving anything at all on turn one — → empty-source
      ok    moving a marble off the board — → out-of-range
      ok    placing onto an occupied square — → occupied
      ok    placing onto the square the slid marble just landed on — → occupied
      ok    placing off the board — → out-of-range
      ok    placing nowhere at all — → out-of-range
      ok    a turn that is not shaped like a turn — → bad-turn
      ok    a move that is not shaped like a move — → bad-turn
      ok    playing out of turn — → not-your-turn
      ok    playing as somebody who is not in the game — → unknown-slot
      ok    playing after the game is over — → finished
      ok    placing onto the square the slid marble just VACATED is legal — slide 5→6, place on 5
      ok    declining step 1 is legal, and turn one has nothing to decline
      ok    legalMoves only ever offers the opponent's marbles, orthogonally — 5 slides offered, all playable
      ok    the reducer routes PLAY TURN and refuses anything else
      ok    createState throws on a bad roster; resetState hands the winner the opening — 5 bad rosters rejected

    Self-play — invariants over whole games
      ok    2000 random games hold every invariant — 2000 games · 1756 won, 189 simultaneous draws, 55 exhausted draws · longest 16 turns, most presses 6
      ok    string slots work as well as integer ones — ended won by black

    Purity — the engine imports nothing but itself
      ok    no React, Prisma, realtime core, alias or third-party import — 10 files, relative sibling imports only

    Race Condition engine: all checks passed.
    exit 0

    $ npm run lint
    ✔ No ESLint warnings or errors

`npm run build` was not run: nothing imports `_lib` until #143, so the build does not exercise a line of this.

### Proving the script can go RED

A verification script nobody has watched fail is a decoration. I committed the engine first, then mutated it one named file at a time (`git checkout HEAD -- <file>` to restore, never `git stash`), with an anchor check that each mutation actually applied so "changed nothing" could not read as "harmless":

| Mutation | Result |
|---|---|
| Orbit spins clockwise instead of counter-clockwise | **RED** — 4 checks |
| The win is read off the board *before* the press | **RED** — 2 checks |
| A simultaneous line is broken into a win | **RED** — 2 checks |
| Step 1 allowed to move your own marble | **RED** — 1 check |
| Adjacency trusts flat-index arithmetic (row wrap) | **RED** — 1 check |
| The win scan forgets the diagonals | **RED** — 3 checks |
| The win scan reports only the first line it finds | **RED** — 2 checks |
| `OVERTIME_SPINS` cut from 5 to 4 | **initially GREEN — see below** |
| A finished game hands the turn to the loser | **initially GREEN — see below** |

The last two were real holes in the harness and are fixed in the second commit:

- The overtime assertions were phrased `OVERTIME_SPINS + 1`, so shortening the rulebook's five extra presses moved the expectation with it. An assertion that cannot fail. The press counts are literal `6` now, and a new first check pins the board size, marble count and overtime count to the published numbers.
- Nothing asserted that a finished board names the player who *ended* it rather than the winner. It now fails both the lose-by-your-own-hand check and self-play.

Both go RED after the fix. The working tree was clean after every probe.

## Notes for the reviewer

- **One export beyond the plan's list.** The plan spells out the state's `drawReason` as the string literals `"simultaneous"` / `"exhausted"` but does not name a constant. I added a frozen `DrawReason` to `constants.js` and exported it, matching the `MoveError` idiom, so #143 has a name to read instead of a literal. The literals are unchanged. Say the word and I will inline them.
- **Slot `0` is exercised on purpose.** The realtime core numbers seats from zero, and an engine that tests a square with `if (cells[i])` treats every one of player 0's marbles as an empty square. Every internal check is `== null` / `!= null` / `===`, and the fixtures use slots `0` and `1` throughout, with one self-play round on string slots to prove neither is assumed.
- **`BadTurn` vs `OutOfRange` is a judgement call.** `BadTurn` covers a payload that is not shaped like a turn (`turn` not an object, `move` not an object); a `place` or `move.from`/`move.to` that is merely not an integer in 0…15 gets `OutOfRange`. So `{ place: undefined }` reports `out-of-range` rather than `bad-turn`. Defensible either way; both are refusals with a message and neither throws.
- **`cellsBeforeOrbit` is stored, and it is a sixteenth array per turn.** The plan makes it the animation contract for #143 (render the pre-press position, then call `orbit` yourself `spins` times). It also round-trips through the room's MySQL JSON column in #144. It is 16 nullable scalars, so the cost is small, but it is a real addition to the persisted state and worth a look now rather than after #144 ships.
- **Overtime fixtures are searched, not hand-built.** The two overtime checks shuffle seeded full boards until they find one matching the shape they need (all six presses line-free; first line on press 3). The seeds are fixed, so the run is deterministic, and both checks anchor on the board they found actually having that property before asserting the engine's answer. If the search ever failed it would report "no board turned up" rather than passing quietly.
- **The `_lib` purity check is a regex over `from "…"` specifiers.** It catches every import in the engine today. A dynamic `import()` or a `require` would slip past it. That seemed proportionate for a directory this size; happy to tighten it.
- The `next lint` deprecation notice ("`next lint` is deprecated and will be removed in Next.js 16") is pre-existing and unrelated to this change.


csalinas-dev:
## 🤖 Review: APPROVED

Checked: security, correctness, efficiency, fit against the plan. The whole risk on this one is *"is it actually Orbito"*, so I went to the publisher's rulebook myself and then wrote a second engine to argue with this one.

### I re-extracted the rulebook independently

I downloaded `230308-FlexiQ-FXG-502-Orbito-Game-Rules-A4-UK-HR-v1.pdf` from flexiqgames.com and inflated its FlateDecode streams with my own Node extractor rather than trusting the plan's quotations. Every sentence the plan and the code quote is in the PDF, verbatim:

- `MOVE ONE OF YOUR OPPONENT'S MARBLES TO AN ADJACENT, EMPTY SQUARE.` — optional, "not possible during the first turn", "only horizontally or vertically, not diagonally".
- `PLACE ONE OF YOUR OWN MARBLES ONTO ANY FREE POSITION` / "Once placed, you cannot move your own marbles anymore. Only your opponent can!"
- "This will cause **ALL** marbles to shift 1 position" / "Both the inner & outer orbits rotate counter-clockwise."
- "Only after pressing the Orbito-button, a valid 4-in-a-row can be created. So never during a player's turn!"
- "a player creates a sequence of 4 in their **opponent's** colour and makes their opponent win!"
- "If, after pressing the Orbito-button, both players simultaneously have a sequence of 4, it's also a draw!"
- "the Orbito-button needs to be pressed 5 more times. The first player to get a sequence wins the game. If there is no sequence after 5 times, it's a draw!"

**The architect's correction stands and the PR built it**: the issue summary's "any single marble" is wrong; the rulebook says opponent's marble, and `moves.js` refuses your own with `MoveError.OwnMarble`. Confirmed at the source, not taken on trust.

I also re-derived "counter-clockwise" from scratch rather than checking the literals against themselves: on screen with row 0 on top, a marble on the top edge moves left, the left edge moves down, the bottom edge moves right, the right edge moves up. That produces `[0,4,8,12,13,14,15,11,7,3,2,1]` and `[5,9,10,6]` and the permutation `[1,2,3,7,0,6,10,11,4,5,9,15,8,12,13,14]` — exactly what `ringTrack` derives and what the script asserts. The tracks really are a grid walk (one `ringTrack(inset)` for both rings), not the literals rearranged; the literals live only in the assertion.

### An independently written shadow engine agrees on every turn

I wrote a second Orbito from the rulebook text — its own orbit built from `(row, col)` edge rules rather than index tracks, its own ten lines, its own turn resolution — and ran it lock-step against `playTurn` for **4000 random games**: board after every press, `spins`, `cellsBeforeOrbit`, `winner`, `draw`, `drawReason`, `winningLines`, the `legalMoves` set, and turn order.

    4000 games shadowed by an independently written engine ·
      3567 wins, 366 simultaneous draws, 67 exhausted draws
    shadow engine: identical on every turn

All three terminal outcomes are exercised. To prove the shadow is not degenerate I flipped `orbit.js` to clockwise and it diverged on game 0, then restored the file.

### The verification script can fail — I broke it 12 ways

Twelve mutations, one named file at a time, `git checkout HEAD -- <file>` to restore (no stash, no directory), each anchored so a no-op replacement reports rather than passes. **Eleven went RED.** Deliberately mutations the PR body does *not* list:

| Mutation | Result |
|---|---|
| overtime runs 6 extra presses instead of 5 | RED (2 checks) |
| `cellsBeforeOrbit` snapshots the board *after* the press | RED (4) |
| `playTurn` mutates the caller's board in place | RED (2) |
| the mover always wins, never the line's owner | RED (4) |
| the overtime guard forgets to require a full board | RED (4) |
| adjacency trusts flat-index arithmetic | RED (1) |
| the orbit runs clockwise | RED (4) |
| an empty line counts as a completed one | RED (7) |
| the turn never passes to the other player | RED (3) |
| the placement square is not checked for being empty | RED (2) |
| a simultaneous draw hands the win to the first line's owner | RED (2) |
| `getResult` builds `hands` via an object round-trip | GREEN — but see below |

The one GREEN was **my** bad mutation, not a harness hole: `Object.values(Object.fromEntries(slots.map(...)))` is semantically identical to `slots.map(...)` for both integer and string slots, so there was nothing to catch. The worktree was clean before and after every probe.

The `OVERTIME_SPINS + 1` fix in the second commit is real — the assertions now read literal `6` and `3`, and cutting the constant goes RED.

### My own probes, on top of the committed script

    ok  playTurn does not mutate the state it was given
    ok  playTurn does not mutate the cells array it was given
    ok  slot 0 can win — winner=0
    ok  getResult reports slot 0 as the winner (not null)
    ok  resetState gives slot 0 the opening after slot 0 won
    ok  reducer resolves { slot: 0 } / a bare 0 / refuses a missing player
    ok  no malformed payload throws or is accepted — 15 refused
    ok  every refusal returns the same state reference
    ok  state survives JSON round-trips and keeps playing
    ok  exhausted draw ran exactly 6 presses, no line at any of them
    ok  every export the plan names is present — 33 exports
    ok  legalMoves is exactly the set canMove allows — 300 games, 0 mismatches

Plus `node .agent/scripts/verify-race-condition.mjs` → 44 checks green, exit 0, and `npm run lint` → "No ESLint warnings or errors" (after `npm install` in a fresh worktree).

### The three things you flagged

- **`DrawReason` beyond the plan's export list — keep it.** Frozen, matches the `MoveError` idiom, literals unchanged, and it gives #143 a name to read instead of a bare `"simultaneous"` in a component. That is strictly better than the plan's shape, not scope creep. `ORBIT_TO` / `TRACKS` / `LINES` / `linesFor` are likewise extras and likewise harmless.
- **`BadTurn` vs `OutOfRange` — the judgement call is fine.** `{ place: undefined }` reporting `out-of-range` ("That square is not on this board.") is only reachable from a caller bug, never from a board sending the square a player clicked. Both refuse, neither throws, both carry a sentence a player can read. Not worth a round trip.
- **Slot `0` — verified, no truthiness bug anywhere.** I checked every read: `isEmpty` / `emptyCells` / `isBoardFull` / `completedLines` use `== null` / `!= null`, `marblesOnBoard` uses `===`, and every fallback is `??` and not `||` (`winner ?? null`, `first ?? slots[0]`, `outcome?.winner ?? null`). Confirmed end to end with slot 0 placing, winning, being reported by `getResult`, and opening the rematch.

### Acceptance criteria

- [x] `_lib/` exports the full list — 33 exports, checked programmatically
- [x] Imports nothing outside `_lib/` — purity check plus reading all 10 files
- [x] `board.js` / `orbit.js` convention headers; `index.js` state shape and pointers
- [x] Opponent's marble only, orthogonal, onto an empty square — own marble refused with `MoveError.OwnMarble`
- [x] A line only wins after the press — shadow-engine agreement plus the dedicated anchored check
- [x] You can hand the win to your opponent — `settle` reads the winner off the board, never assumes the mover
- [x] Simultaneous is a draw, never broken, `winningLines` carries both colours
- [x] Overtime resolves inside the same call, stops at the first line, otherwise `exhausted` with `spins === 6`
- [x] A refusal returns the same state reference and never throws; `createState` throws on a bad roster
- [x] Verify script passes and its output is quoted in the PR body
- [x] `npm run lint` clean
- [x] `CLAUDE.md` documents the engine

### Non-blocking observations

- `LINE_LENGTH` is exported and pinned by the constants check but nothing derives from it — `lines.js` builds from `SIZE`. Harmless on a 4x4 board where they are equal; just know it is documentation, not a lever.
- Nothing in the committed script asserts that a *successful* `playTurn` leaves the previous state's `cells` untouched. It does — I checked — and it will matter for React identity in #143, so it is worth an assertion when that lands.
- The purity check's `from "..."` regex would miss a single-quoted specifier or a dynamic `import()`. Prettier and ESLint enforce double quotes here, so it holds today; leave it.
- `cellsBeforeOrbit` persisting into the room's MySQL JSON column in #144 is 16 nullable scalars per row. Fine, and it is the animation contract — flagged only so #144 does not rediscover it.

Nothing blocking. This is the rulebook, correctly encoded.

Issue #138, opened 2026-08-10
fix: body { height: 100vh } makes every page ~86px taller than the visible area on iOS Safari


csalinas-dev:
Split out of #136 / #137 at the reviewer's request, so the analysis is not lost with that PR. **This is not the Tic-Tac-Overflow board bug** — that one was the glyph's percentage height and is fixed in #137. This is the separate, site-wide issue found while investigating it.

## What

`src/app/globals.css:69` sets:

    body {
      height: 100vh;
    }

On iOS Safari `100vh` resolves to the **large** viewport — the height the page would have with the browser chrome retracted — not the height you can currently see. So `body` is taller than the visible area by the height of the collapsed toolbars (~86px on an iPhone 16 Pro Max at 440x956), and every page on the site scrolls slightly when it looks like it should not.

## Why it is filed separately

It is real, it is reproducible, and it affects **every page**, not just the game boards. It was explicitly ruled out as the cause of #136 — the board's geometry contains no viewport unit in its block axis, and the reported screenshot showed the footer on screen with room to spare, so the 427-vs-392 board there came from somewhere else entirely.

It was left out of #137's scope deliberately: swapping the unit as a drive-by would change the layout of every page on the site inside a PR whose diff is three CSS declarations on one game.

## Likely fix, not yet verified

`100dvh` (tracks the viewport as the toolbars show and hide) or `100svh` (the small viewport, stable but always the shorter one). Which is right depends on whether the layout should resize as the user scrolls, and it wants checking against the pages that actually depend on a full-height `body` — the games in particular, since `Gameboard.jsx` already budgets against `100svh`.

Nobody on the pipeline has a real iPhone, so whatever ships here needs either a device check or a correctness argument that does not rest on Chrome measurements.

---

## 🤖 Implementation Plan

_Written by the architect agent on 2026-08-10 08:39 UTC. The implementor builds exactly this._

### Goal

Nothing on this site claims to be "as tall as the screen" using a unit that, on
iOS Safari, means *taller than the screen*. After this ships, `body` — and every
element that budgets against the full viewport height — is at most as tall as the
area the user can actually see, in every browser-chrome state, so the ~86px of
phantom scroll on every page is gone. On every engine without retractable
browser chrome (all desktop browsers, and the headless Chrome this pipeline
measures with) the change is a **provable no-op**: nothing moves by a single
pixel.

### The defect, restated precisely

[...]

PR #146, opened 2026-08-15
feat(RC): animated Orbito board and hotseat play

### LINKED ISSUE #143: feat(RC): animated Orbito board and hotseat play

csalinas-dev:
Closes #143

Part 2 of #141. `/games/race-condition` becomes a playable, animated Orbito board
for two people on one device. Zero network calls in the tree — #144 owns all of
that, and the context shape it will fill is already here.

## What changed

Read in this order.

- **`_lib/moves.js` + `_lib/playTurn.js` + `_lib/index.js`** — the one engine
  change in scope: step 1 is extracted as `applyMove(cells, move)` and `playTurn`
  calls it. The board needs that post-slide position twice (the staged preview,
  and the set of legal placements) and re-deriving it in a component would be
  re-implementing the rule. It returns `cells` **by reference** when the move is
  declined, so `playTurn`'s existing `played === cells ? [...cells] : played` is
  unchanged and still correct.
- **`players.js`** — cyan/gold, structured exactly like `connect-404/players.js`,
  plus `opponentOf` (step 1 is about the opponent, so most sentences need them by
  name). Every marble carries its player's initial; colour is never the only
  signal.
- **`context/`** — `{ state, dispatch, canPlay, canReset, online }`. The module
  comment states the rule #144 depends on: `canPlay` is the only authority on
  whether this device may act, `online?.youSlot` the only source of which seat it
  holds, and nothing below may re-derive either from `game.turn`.
- **`board/ids.js`** — synthetic marble ids. Presentation only, no rules: it keeps
  one DOM node attached to one marble so a press is a slide instead of a repaint.
  The permutation is `ORBIT_TO` from `_lib`, not a rotation worked out here.
- **`hooks/useOrbitAnimation.js`** — the playback. Frame 0 is
  `lastTurn.cellsBeforeOrbit`, held 420ms; each later frame is `orbit()` applied
  to the one before it, `lastTurn.spins` times (460ms for the first press, 300ms
  for the overtime ones). The last frame **is** `state.cells` by construction.
- **`hooks/useTurnDraft.js`** — the staged turn. `legalMoves` / `applyMove` /
  `emptyCells` from `_lib` decide everything it offers. Placing is the only
  commit, and placing without staging a slide *is* declining step 1, so there is
  no skip button.
- **`hooks/useCellNavigation.js`** — roving tab stop, arrows clamped at the rim,
  every square reachable playable or not.
- **`board/`** — `Cell` (button + socket + rings, `aria-disabled` never
  `disabled`), `Marble` (positioner owns `transform`, disc owns the scale
  animations), `MarbleLayer`, `OrbitIndicator`, `Board`. `Board` owns no game
  state and takes `overlay` for #144's curtain.
- **`components/`, `Toolbar.jsx`, `Game.jsx`, `RaceCondition.jsx`, `page.js`** —
  ported from Connect 404 with this game's copy. `Endgame` says who won *and*
  whether the mover turned the board into their opponent's line.
- **`CLAUDE.md`** — three bullets: the marble-layer technique and id ordering, the
  animation contract and the two traps in it, and the staged turn.

### Two things worth a second look

**The double-applied slide.** `useOrbitAnimation`'s ids are always the
**committed** ids; `Game` layers the staged slide on at render time. Fold it into
the hook instead and the commit's frame 0 applies the same slide a second time
from a square the first application already emptied, and the marble disappears.
The structure is the defence, not a special case.

**Marbles render in id order, not board order.** Ids only increase, so a list
sorted by id never reorders. React reordering keyed children calls `insertBefore`,
which is a remove and a re-insert — that cancels the transition mid-flight.

## How it was verified

Checks:

    $ node .agent/scripts/verify-race-condition.mjs   → all checks passed
    $ npm run lint                                    → ✔ No ESLint warnings or errors
    $ npm run build                                   → ✓ Compiled successfully; /games/race-condition ○ (static)

`npm install` and `npx prisma generate` are needed first in a fresh worktree; the
build fails at the API routes without the generated client, unrelated to this PR.

**The extraction's guard bites.** With `applyMove` deliberately reversed
(`next[move.from] = next[move.to]`), the verify script fails 3 checks and exits 1
— "placing onto the square the slid marble just landed on", "…just VACATED is
legal", and the 2000-game self-play. Reverted, it is green again.

**The orbit was measured in a browser, not reasoned about.** A probe drove the
real page through headless Chrome, held on to the marbles' DOM nodes across a
press, and compared where they landed against a counter-clockwise walk
**re-derived inside the probe** from the rulebook's words rather than copied from
`_lib`:

    marblesMeasured: 2
    midFlight:  [{from: 8, to: 12, progress: 0.766}, {from: 9, to: 10, progress: 0.761}]
    moved:      [{from: 8, expected: 12, actual: 12, stillTheSameNode: true},
                 {from: 9, expected: 10, actual: 10, stillTheSameNode: true}]
    everyMarbleLandedWhereCounterClockwiseSays: true
    everyMarbleWasInTransit: true

`progress` strictly between 0 and 1 is a marble part-way between two squares, and
`stillTheSameNode` is the node-identity claim. Both rings are covered (8→12 is the
outer track, 9→10 the inner). Flipping the probe's own derivation to clockwise
makes it report `false` with `expected: 4, actual: 12` — so it discriminates, and
it is not passing because it agrees with itself. A frame-by-frame sample of the
same turn:

    124ms  m0@8  m1@9  new@2      <- the position the player built, held
    378ms  m0@8  m1@9  new@2
    499ms  in transit (--row/--col at the destination, still between squares)
    624ms  m0@12 m1@10 new@1      <- settled, and it does not move again

**A whole game, and the rules that are easy to get wrong.** Another probe played
turn by turn:

    turn one:   sources: [], placeable: 16, "Place a marble."     (nothing to slide)
    turn two:   sourcesOfferedToGold: [9]  — only Cyan's marble, label
                "Row 3, column 2, Cyan's marble. Move this marble"
    picked up:  destinations: [5, 8, 10, 13]  (orthogonal, empty)
    staged:     vacatedIsPlaceable: true  — "The square your slide emptied…"
                landedIsPlaceable: false
    own marble: tapped, marbles 2 → 2, alert "That marble cannot be slid this
                turn. You may only slide Gold's marbles…"
    mid-orbit:  tapped, marbles 3 → 3  (the board is closed while it turns)
    overtime:   PRESS 1 OF 6 … PRESS 6 OF 6, then a draw, "The board filled and
                five more presses found nothing." Play again offered.
    a win:      winning cells [3, 6, 9, 12], 7 other marbles dimmed, "Cyan wins /
                Four in a row after the orbit." and the announcer names the line.

**Reduced motion.** Driven with `Emulation.setEmulatedMedia`
(`prefers-reduced-motion: reduce`): 120ms after a commit every marble is already
on a square (< 3px off centre), at cells 12/10/1 — the identical position the
animated run settles to. Same board, no trip.

**Keyboard.** One tab stop on the board; `0 →ArrowRight→ 1 →ArrowDown→ 5
→ArrowLeft→ 4 →ArrowUp→ 0`, clamped at the rim, `Home`/`End` to the ends of the
row; all **16** squares reached by walking; an occupied square takes focus and is
`aria-disabled`, not `disabled`.

**Screenshots** (`shot.mjs`, viewport read back from the page — `viewport.w`
was 440 and 1280, matching what was requested, so this is not #134 again):

    node .agent/scripts/shot.mjs http://localhost:3143/games/race-condition 440  rc-440.png
    node .agent/scripts/shot.mjs http://localhost:3143/games/race-condition 1280 rc-1280.png

Both report `scrollWidth === innerWidth` (no horizontal scroll) and the board's
computed heights are px from `max-width` + `aspect-ratio` — no `vh`/`svh`/`dvh` in
a block axis (#138). They are in the worktree, untracked:

    .../csalinas-dev_csalinas-dev-site__impl-143/rc-440.png       mid-game, a marble picked up
    .../csalinas-dev_csalinas-dev-site__impl-143/rc-1280.png      the same, wide
    .../csalinas-dev_csalinas-dev-site__impl-143/rc-mid-orbit.png six marbles caught between squares

The mid-orbit one is a genuine frame of the animation, with one stylesheet
override injected by the probe (`transition-duration: 6s` on the marble layer) so
a still camera could catch it. The game's own timings are untouched.

## Notes for the reviewer

- **A bug the probe caught, and how.** My first version restored the "last watched
  turn" ref in the animation effect's cleanup, meaning to be safe under React's
  development double-invoke. That cleanup runs immediately *before* the next
  turn's effect, so every turn after the first looked like the first: settled on
  arrival, animation silently skipped. It looked completely fine by eye — the
  board was always correct — and only the frame-by-frame probe showed the marbles
  already at their final squares 124ms after the click, with every DOM node
  replaced. The cleanup now cancels timers and nothing else, and the comment says
  why. This is the strongest argument I have for the probe over eyeballing it.
- **Two cases I did not observe, only implemented.** The "mover orbited their
  opponent into a line" endgame copy (`winner.slot !== mover.slot`) and the
  simultaneous draw with both colours' lines shown. Both are data-dependent — the
  engine's own self-play puts simultaneous draws at 189/2000 — and I did not find
  a way to force either from the UI in reasonable time. The exhausted draw and an
  ordinary win are both shown above.
- **Where the winning colour lives.** The plan asks for each line's ring coloured
  by `cells[line[0]]`. The ring on the cell is neutral (`--foreground`) and the
  *marble* carries the owner's colour as a drop-shadow glow, so a two-colour draw
  still shows each line in its own colour. Say the word if you want the ring
  itself tinted.
- **One rule the plan did not name.** Once a slide is staged, tapping the marble
  where it now stands takes it back. Without it a mis-tapped slide could only be
  committed, and the plan's "tap again to cancel" only covers a marble picked up
  and not yet moved. Sources are also hidden while a slide is staged, because
  `legalMoves` is computed against `game.cells` and would otherwise light up the
  square the staged marble has already left.
- **Deliberately not here**, per Out of scope: the mode picker, the lobby,
  `multiplayer.js`, `online/`, `components/Panel.jsx`, `components/Button.jsx`,
  the games-index card and the artwork. `overlay` is already a `Board` prop with
  nothing passing it.


csalinas-dev:

Checked out `pr/146`, diffed against `epic/race-condition` (2 commits, 30 files), installed,
built, and drove the real page in headless Chrome. One blocking bug, in exactly the case the
PR body flagged as implemented-but-unobserved.

### Blocking

1. **`components/TurnBanner.jsx:83-87` — the turn banner names the LOSER as the winner whenever the mover orbited their opponent into a line.**

```js
const headline = game.draw
  ? "Draw"
  : game.winner !== null
    ? `${player.name} wins`      // <- player is playerFor(players, game.turn)
    : `${player.name} to move`;
```

`player` is `mover` from `Game.jsx:95` — `playerFor(players, game.turn)`. The engine
deliberately leaves `state.turn` with the player who **ended** the game (that is
`lastTurn.by`), which is *not* the winner when the winning line was completed by the press
rather than by the placement. The banner then also takes that player's colour
(`TurnBanner.jsx:111`, `--slot`) and their initial in the chip (`:113`).

**Concrete inputs.** A real 16-turn game replayed by clicking cells on
`/games/race-condition` (turn list below). The final board matches `state.cells` exactly,
`winningLines` is `[[3,6,9,12]]`, `winner` is `p1` (Cyan), `lastTurn.by` is `p2` (Gold).

**Concrete wrong outcome.** The rendered page says both of these at once:

- above the board, large, gold chip, gold rule: **"Gold wins — Four in a row, after the press."**
- below the board: **"Cyan wins — Gold turned the board into Cyan's line. Cyan opens the next game."**

The highlighted line on the board is Cyan's. `Endgame`, `Announcer` ("Cyan wins, on Gold's
press…") and the ring highlight are all correct; only the banner is wrong, and it is the
loudest element on the screen.

**This is not an edge case.** 3000 self-played games through `_lib`: 2655 wins, of which
**1022 (38.5%) have `winner !== lastTurn.by`**. More than a third of all decided games render
a banner that contradicts the panel eight lines below it.

**Fix.** Read the winner off `game.winner`, not `game.turn` — pass a `winner` descriptor into
`TurnBanner` (or `playerFor(players, game.winner)`) and use it for the headline, the chip and
`--slot`. `game.turn` stays right for the "to move" branch. While you are there, `beat()` at
`:96` returns "Four in a row, after the press." for every win; the orbited-in case deserves
the same sentence `Endgame` already gets right, and at minimum must not contradict it.

Repro turn list (each entry is one committed turn; click `move.from`, `move.to`, then `place`):

```json
[{"move":null,"place":5},{"move":null,"place":5},{"move":{"from":9,"to":8},"place":14},
 {"move":null,"place":10},{"move":null,"place":14},{"move":{"from":9,"to":13},"place":11},
 {"move":null,"place":2},{"move":{"from":1,"to":5},"place":12},{"move":{"from":13,"to":14},"place":5},
 {"move":null,"place":4},{"move":{"from":9,"to":13},"place":5},{"move":{"from":9,"to":13},"place":11},
 {"move":null,"place":10},{"move":null,"place":1},{"move":null,"place":3},{"move":{"from":9,"to":10},"place":9}]
```

### Non-blocking

- **`board/Board.jsx:69-73` — "Place Cyan marble here".** `labelFor` builds
  `Place ${you?.name ?? "your"} marble here`. The `"your"` fallback reads correctly but never
  fires in hotseat, so every placeable square (usually a dozen at a time) announces the
  ungrammatical form. Verified in the DOM: "Row 2, column 2, empty. The square your slide
  emptied. Place Cyan marble here". Want `${you.name}'s` or just "your".
- **`components/TurnBanner.jsx:113` — the draw chip keeps a player's colour.** `.draw`
  neutralises the container's `border-left` and forces the headline to `--foreground`, but the
  `Chip` still takes `background-color: var(--slot)`, which is the last mover's colour.
  Observed cyan on the simultaneous draw and gold on the exhausted one. The component comment
  says the banner "does not get to imply otherwise by leaving the last mover's name up in
  their colour" — the chip is the one bit that still does.
- **`board/Board.jsx:151-159` — the refusal alert explains the wrong rule once a slide is
  staged.** `draft.sources` goes empty when `move !== null` (`useTurnDraft.js:64-67`), so
  tapping a *different* opponent marble after staging a slide is inert and fires "You may only
  slide {opponent}'s marbles, and only onto an empty square beside one." The real reason is
  "you have already used your one slide this turn". Screen-reader-only, low severity, but it
  is a wrong explanation rather than a missing one.
- `Announcer` nests `list()` inside `list()` on the simultaneous draw, producing "…row 1,
  column 3 **and** row 1, column 4 **and** Gold at…". Understandable, slightly clumsy.

### Verified — everything below is correct and I am not asking for changes

**The orbit direction, measured from the rendered DOM rather than read off an import.**
Sampled the marble layer's `--row`/`--col` frame by frame through a 6-press overtime sequence
and recomputed each frame independently from `ORBIT_TO`. Frame 0 `GCCGGGCCCCGGCGGC` →
observed frame 1 `CCGCGCGGGGCCCCGG`, which is exactly `next[ORBIT_TO[i]] = old[i]` (0→4, 8→12,
15→11, 5→9, 10→6 …). All 8 rendered states of that turn matched. Three separate full games
(16, 15 and 16 turns) ended with the rendered board **identical to `state.cells`** — a
mirrored renderer would have diverged by turn 2. `board/ids.js:38-46` uses the same expression
as `_lib/orbit.js:118-126`, so cells and ids cannot drift. `OrbitIndicator`'s four chevrons
also point the right way (left→down, bottom→right, right→up, top→left).

**No marble vanishes or doubles.** Replayed a 15-turn game in which **13 turns stage a
slide**, sampling the marble layer at ~70ms through every staging *and* every commit
animation: zero count changes, zero duplicate squares, zero marbles left on a vacated square.
Also confirmed per turn that the staged preview empties `move.from`, fills `move.to`, and that
tapping the marble where it now stands takes the slide back cleanly. The
committed-ids/layer-at-render-time structure holds.

**The animation actually runs on turns 2, 3 and 4** (per the note about the
cleanup-restores-ref false green). Each of those turns rendered three distinct states —
settled → `cellsBeforeOrbit` held → post-orbit — with the pre-existing marbles' DOM nodes
surviving (probe-set `data-` tags intact) and only the newly placed marble mounting. The
`const teardown = cancel` fix is doing its job.

**Clicking during the animation does nothing.** Clicked 4 cells mid-sequence on every turn of
two full games, and mid-overtime: marble count never moved.

**Overtime.** Caption went `Orbit → Press 1 of 6 → … → Press 6 of 6 → Orbit`, all six presses
animated in order, none skipped, board closed throughout, final board identical to the engine.

**Reduced motion.** Driven with `Emulation.setEmulatedMedia`: media query matches, marble
`transition-duration` is `0s`, and the board is at its settled position within 40ms on turn 1
**and turn 2** — the same squares the animated run reaches.

**Interrupted sequence.** Pressed Restart 300ms into a commit (mid frame-0 hold): timers
cancelled, board empties, no stale ids, and the next game animates normally.

**Rules.** Turn one offers no slide and 16 placements. Turn two offers only the opponent's
marbles (label "Row 3, column 2, Cyan's marble. Move this marble"). Tapping your own marble
selects nothing and raises the refusal alert. With a slide staged, the vacated square carries
`vacated` + is placeable and `move.to` is not.

**Simultaneous draw**, forced by replaying a searched game since the UI cannot be steered to
one: banner "Draw / Both lines completed on the same press", panel "…A tie is a tie", both
lines ringed (cells 0-7) with each marble glowing its own colour, announcer names both lines
by owner. Correct — this half of the disclosed pair is fine; only the win half is not.

**The `applyMove` extraction, and its guard.** `node .agent/scripts/verify-race-condition.mjs`
→ all checks passed, exit 0. Then I mutated `_lib/moves.js` (`next[move.to] = next[move.from]`
→ `next[move.to] = null`, with an anchor assertion so a silent no-op could not read as a pass)
and re-ran: **3 checks FAILED, exit 1**. Restored via `git checkout HEAD -- <file>`, green
again. The script is real evidence, not a decoration.

**Gate.** `npm run lint` → no warnings or errors. `npm run build` → compiled successfully,
`/games/race-condition` ○ (static), 12.4 kB.

**Viewport.** `shot.mjs` reported `viewport.w` of **440** and **1280**, both matching what was
requested, so this is not #134. `scrollWidth === innerWidth` at 440. No `vh`/`svh`/`dvh`
anywhere in the game directory; the board is `max-width: 26rem` + `aspect-ratio: 1/1`.

**Keyboard.** One tab stop on the board; `0 →Right→ 1 →Down→ 5 →Left→ 4 →Up→ 0`, clamped at
the rim, `End→3`, `Home→0`. No cell carries native `disabled`.

**Rules-in-components grep.** Every hit of
`ORBIT_TO|orbit(|areAdjacent|completedLines|LINES|cellIndex|cellCoords` outside `_lib/` is an
import from `_lib` or a call into it.

**Security / scope.** No network calls, no new dependencies, no `dangerouslySetInnerHTML`, no
untrusted input. Scope matches the plan: nothing under `src/lib/realtime/`, no
`multiplayer.js`, no `online/`, no `games.js`, no artwork. `overlay` is a `Board` prop with
nothing passing it.

### Acceptance criteria

- [x] Route renders, playable end to end, wins and both draws reported
- [x] One `PLAY TURN` dispatch per turn; nothing half-finished in `state.game`
- [x] Slide offers only the opponent's marbles, orthogonally, onto empty squares
- [x] Declining is always possible; turn one offers no slide
- [x] Vacated square placeable, `move.to` not
- [x] Orbit animates, nothing teleports or cross-fades
- [x] Overtime animates all 6 presses with the counter; board closed until settled
- [x] Final frame identical to `state.cells`; reduced motion lands in the same place
- [x] Simultaneous draw shows both lines, never broken
- [ ] **Win reported correctly, including mover-orbits-opponent-into-a-line** — the line and
      the panel are right, the banner is not (blocking #1)
- [x] Every rule decision traces to `_lib`
- [x] Context is the five-key shape; `online` null; seat only from `online?.youSlot`
- [x] Keyboard: one tab stop, arrows walk all 16, every cell reachable
- [x] Screen reader gets the staged slide, the committed turn and the result
- [x] 440px / 1280px, no horizontal scroll, no viewport units in a block axis
- [x] Verify script passes after the extraction — and proven to fail without it
- [x] Lint and build clean

One line in `TurnBanner` and this is an approve.

csalinas-dev:
## Addressed — `f21b4dd`

### Blocking: the banner named the loser

Fixed at the source. `Endgame`, `Announcer` and the ring were all right because
they read the winner off the **board** — `result.winner` / `game.winner` — and the
banner was reading `game.turn`. Rather than give the banner a fourth correct
derivation, `Game.jsx` now makes **one**:

```js
const outcome = useMemo(() => {
  const winner = result.winner === null ? null : playerFor(players, result.winner);
  const lastMover = game.lastTurn === null ? null : playerFor(players, game.lastTurn.by);

  return { winner, lastMover, orbitedIn: winner !== null && lastMover !== null && winner.slot !== lastMover.slot };
}, [game.lastTurn, players, result.winner]);
```

`TurnBanner` and `Endgame` are handed that same object. The banner takes the
headline, the chip initial and `--slot` from `outcome.winner ?? player`, so the
name, the colour and the letter move together and `game.turn` is left doing the
one job it is right for — the "to move" branch. `Endgame` no longer computes
`orbitedIn` for itself; it destructures it. `beat()` now returns the same
orbited-in sentence the panel gets, from the same flag, so the two cannot
disagree by construction rather than by both being edited carefully.

### Making the case reachable — the part that mattered

You were right that this shipped because I could not get it on screen. New
committed script, `.agent/scripts/rc-ending.mjs`:

```
node .agent/scripts/rc-ending.mjs <orbited-in|simultaneous|exhausted> [--seed N] [--eval OUT.js]
```

It seeded-self-plays the **real engine** (every slide from `legalMoves`, every
placement from `emptyCells` of the post-slide board — so every turn it picks is a
turn a human could have clicked) until it finds the ending you asked for, then
re-runs the turn list through the engine as an anchor: a list that does not replay
to the ending it was found for exits 1 rather than producing a screenshot of a
different game. With `--eval` it writes a page script for `shot.mjs` that replays
the game **by clicking the real board** and reads the result back out of the DOM.

Repeat it with:

```bash
node .agent/scripts/rc-ending.mjs orbited-in --eval /tmp/rc-orbited-in.js
npx next start -p 3143 &
node .agent/scripts/shot.mjs http://localhost:3143/games/race-condition 900 rc.png /tmp/rc-orbited-in.js
```

Seed 4 is a 12-turn orbited-in win, seed 2 a 15-turn simultaneous draw, seed 24 a
16-turn exhausted draw (6 presses) — same games every run.

**Orbited-in win, rendered** (Gold ends it, Cyan wins on the press):

| where | before | now |
|---|---|---|
| banner headline | "Gold wins" (gold chip, gold rule) | **"Cyan wins"** — `rgb(79,193,255)` = `--component`, chip `C` |
| banner sub-line | "Four in a row, after the press." | **"Gold turned the board into Cyan's line."** |
| endgame panel | "Cyan wins — Gold turned the board into Cyan's line." | unchanged, still right |
| announcer | "Cyan wins, on Gold's press…" | unchanged, still right |
| ringed cells | `[0,5,10,15]` | unchanged — Cyan's diagonal, matching `winningLines` |

Four elements, one name.

**Simultaneous draw, rendered**: banner "Draw / Both lines completed on the same
press.", headline `rgb(204,204,204)`, chip background `rgba(0,0,0,0)` with a
`rgba(204,204,204,0.54)` border, panel "…A tie is a tie.", both lines ringed and
each marble glowing its own colour. **Exhausted draw**: "The board filled and
nothing was found.", 6 presses, same neutral chip.

### Non-blocking

- **`"Place Cyan marble here"`** → possessive, via one local so both call sites
  (placeable and vacated) stay in step. DOM now reads `Row 1, column 1, empty.
  Place Cyan's marble here` and `Row 2, column 1, empty. The square your slide
  emptied. Place Cyan's marble here`. The `"your marble"` fallback is kept for the
  spectator case, where it is the grammatical one.
- **The draw chip's colour** → hollow rather than a neutral fill: transparent
  background, `--absentForeground` border and text. `--absentForeground` is
  translucent, so a filled chip could not have carried the dark text the coloured
  one uses. Measured above; `box-sizing: border-box` is global so the chip is
  still 33.6px at 440px and nothing moved.
- **The refusal alert explaining the wrong rule** → the two refusals are two
  different rules and now say so. New `refusalFor(draft, opponent)` in `Board.jsx`,
  and the reason is captured **at the tap** rather than recomputed at render, so
  taking the slide back while the alert is still up cannot rewrite it to explain a
  rule the player never hit. Verified by driving the page:
  - no slide staged, own marble tapped → *"That marble cannot be slid this turn.
    You may only slide Gold's marbles, and only onto an empty square beside one."*
  - slide staged, another marble tapped → *"You have already slid a marble this
    turn. Place one of your own to finish it, or tap the marble you slid to take
    the slide back."*
- **`Announcer` nesting `list()`** → the outer join is `"; "`, so the tie reads
  "…row 4, column 1**;** Cyan at row 1, column 3…" instead of an "and" that sounds
  like a fifth square.

### Gate

- `node .agent/scripts/verify-race-condition.mjs` → all checks passed, exit 0.
  Re-proved it can still go red: mutated `_lib/moves.js:79` to
  `next[move.to] = null`, echoed the line back as an anchor so a silent no-op
  could not read as a pass, re-ran → **exit 1**; `git checkout HEAD -- <file>`,
  green again. (No `_lib` change in this commit — the guard is re-proved, not
  re-earned.)
- `npm run lint` → no warnings or errors. `npm run build` → compiled successfully,
  `/games/race-condition` ○ static, 12.7 kB.
- 440px: `scrollWidth === innerWidth === 440`, board 408px. 1280px clean.
- Nothing you verified was touched: no change to the orbit, `ids.js`,
  `useOrbitAnimation`, `useTurnDraft`'s rules, `applyMove`, or any `_lib` file.
  The diff is five components, one new script and three CLAUDE.md bullets.

### One thing I'd flag myself

`Endgame`'s props changed from `{ mover, winner }` to `{ outcome }`, which is a
component you had already passed. It renders the identical four sentences — the
only behavioural change in that file is that `orbitedIn` arrives instead of being
recomputed — but it is a file you signed off on, so it is worth a second look
rather than a skim.

csalinas-dev:
## 🤖 Review: APPROVED

Re-review of the revision `f21b4dd`. Scoped to the fix and the four nits; the earlier pass's findings (orbit direction, no vanishing/doubling marbles, animation on turns 2–4, the `applyMove` guard) were not re-derived.

### 1. The blocking bug — fixed at the source

`Game.jsx` now derives one `outcome` (`{ winner, lastMover, orbitedIn }`) off `result.winner` and hands the *same object* to `TurnBanner` and `Endgame`. `Endgame` no longer computes `orbitedIn`; `TurnBanner` no longer reads a name off `player` for the win case, and `named = winner ?? player` moves the colour, the initial and the chip with it.

Grepped every surviving read of `turn`/`winner` outside `_lib`, and each is doing the job that field is correct for:

| site | read | correct because |
|---|---|---|
| `Game.jsx:58` | `youSlot = online ? online.youSlot : game.turn` | hotseat seat = whoever is up |
| `Game.jsx:95` | `onTheClock = playerFor(players, game.turn)` | "to move", not "wins" |
| `Game.jsx:96` | `opponent = opponentOf(players, game.turn)` | whose marbles may be slid |
| `Announcer.jsx:42` | `upNext = playerFor(players, game.turn)` | who is up next |
| `TurnBanner` | `game.draw` / `game.drawReason` | draw is not in `outcome`; authoritative fields |

No surviving local winner fallback in either consumer. One derivation, one `useMemo`.

### 2. Verified on rendered pages, not in the diff

`.agent/scripts/rc-ending.mjs` reproduces exactly the seeds claimed — `orbited-in` seed 4, `simultaneous` seed 2, `exhausted` seed 24 — and each turn list replays through the engine to the same ending (the script's own anchor). I then replayed all three by clicking the real board via `shot.mjs` against `next start`, plus a fourth game (seed 1) for the ordinary placement win the script does not cover.

**orbited-in (seed 4)** — Gold ended it, Cyan won:
```
banner    "Cyan wins" / "Gold turned the board into Cyan's line."
          chip "C", chipBackground rgb(79,193,255), --slot var(--component)
endgame   "Cyan wins — Gold turned the board into Cyan's line. Cyan opens the next game."
announcer "... Cyan wins, on Gold's press, with a line at row 1, column 1, ..."
ringed    [0,5,10,15]   all four marbles initial "C"
```
Banner, panel, announcer and the highlighted line all name Cyan. That is the bug gone.

**simultaneous (seed 2)** — banner "Draw" with the hollow chip (`chipBackground rgba(0,0,0,0)`, border `rgba(204,204,204,0.54)`, headline `rgb(204,204,204)`), panel "Draw — ... A tie is a tie.", `ringed [0,2,4,6,8,10,12,14]` with both colours' lines lit in their own colour, neither broken.

**exhausted (seed 24)** — 6 presses on the last turn, banner "Draw / The board filled and nothing was found.", panel "…five more presses found nothing.", no rings.

**ordinary win (seed 1)** — "Cyan wins" / "Four in a row, after the press.", `orbitedIn` correctly false, panel agrees.

Also checked `resetState`: "{winner} opens the next game" is true even in the orbited-in case, since `resetState` seeds `first` from `winner`, not `turn`.

### 3. The four nits — spot-checked in the DOM

- Possessive labels: `"Row 3, column 3, empty. Place Cyan's marble here"`, and the vacated square `"Row 1, column 3, empty. The square your slide emptied. Place Cyan's marble here"`.
- Hollow neutral draw chip: computed styles above; screenshot confirms.
- `refusalFor(draft, opponent)` captured at the tap — tapping your own marble with nothing staged gives *"You may only slide Gold's marbles, and only onto an empty square beside one."*; tapping another marble with a slide staged gives *"You have already slid a marble this turn…"*. Two rules, two sentences.
- `"; "` join: `"— Gold at row 1, column 1, … and row 4, column 1; Cyan at row 1, column 3, … and row 4, column 3."` No nested "and".

### 4. Hygiene

- `f21b4dd` touches no `_lib` file (`git diff --stat f21b4dd~1 f21b4dd -- _lib/` is empty).
- `node .agent/scripts/verify-race-condition.mjs` → exit 0, all checks passed. Proved still able to fail: dropping the `next[move.from] = null;` line in `applyMove` (with an anchor assertion that the line was actually found) turns it red — *3 check(s) FAILED* — restored to clean afterwards.
- `npx next lint` → no warnings or errors. `npm run build` → clean (needs `npx prisma generate` first in a fresh worktree).

### Acceptance criteria
- [x] A win is reported with its line(s) highlighted, **including the mover-orbits-opponent-into-a-line case and the both-colours-on-one-press draw** — all four endings rendered and read back out of the DOM.
- [x] Every rule decision traces to a `_lib` call; no rule moved into a component in this commit.
- [x] Verify script green and provably red on a mutation; lint and build clean.

### Non-blocking observations
- `components/Announcer.jsx:58-64` still works out the winner for itself (`game.winner`, and `winner.slot === by` for its own "on Gold's press" phrasing) rather than taking `outcome`. It reads the *correct* field, so it cannot reproduce the original bug, and I confirmed it agrees with the banner and panel in all four endings — but it is the third phrasing of `orbitedIn` in the tree. Worth folding into `outcome` when #144 next touches these components.
- `Game.jsx:127` shadows the new `outcome` memo: `const outcome = draft.selectCell(index)` inside `onCellActivate`. Harmless today (the callback never reads the outer one), but the outer name is now load-bearing enough to deserve not being shadowed.
- `TurnBanner`/`Endgame` destructure `outcome` unguarded, so #144 must keep passing it; a `= {}` default would make that a missing name rather than a crash.
- The banner announces the winner while the board is still turning — `if (winner !== null)` precedes `if (spinning)` in `beat()`, so on a six-press overtime win the headline reads "X wins" about two seconds before the deciding press lands, unlike `Endgame`, which waits for `!anim.spinning`. Pre-existing, not introduced by this commit.

PR #147, opened 2026-08-15
feat(RC): online multiplayer and site integration

### LINKED ISSUE #144: feat(RC): online multiplayer and site integration

csalinas-dev:
Closes #144

Base branch is `epic/race-condition`, not `main`.

## What changed

Read in this order — the first commit is the substance, the rest is assembly.

1. **`hooks/turnKey.js`, `useOrbitAnimation.js`, `useTurnDraft.js`** — the two hazards the plan called out. Both hooks compared by object identity, which is right hotseat and wrong in a room: every snapshot is a fresh `JSON.parse` and the same revision is re-delivered deliberately. They now compare a value key derived from the turn, commented at both call sites.
2. **A third hazard the plan did not have, found by verification** — see "the bug" below. The orbit's timers moved out of the effect cleanup into an unmount-only effect.
3. **The #143 carry-overs** — `Announcer` consumes the shared `outcome` instead of deriving a third phrasing of who won; `Game.jsx`'s tap handler stops shadowing that memo (`staged`); `TurnBanner`/`Endgame` destructure `outcome` defensively.
4. **`multiplayer.js` + one line in `registry.js`** — fixed seats `[0, 1]`, room state *is* engine state, `createState` ignores the roster, `NEW_GAME` handled locally and everything else handed to `_lib`, refusals flattened to a sentence. No rule is restated and `registry.js` is the only file under `src/lib/realtime/` this PR touches.
5. **`online/OnlineGame.jsx`** and the screens around it (`Lobby.jsx`, `RaceCondition.jsx`, `online/OnlineBar.jsx`, `online/RoomCode.jsx`, `components/Panel.jsx`, `components/Button.jsx`). `Game.jsx` gains exactly two props: `hint` and `outcome`.
6. **`.agent/scripts/rc-room.mjs`** — `check` and `park`, plus `park --eval` for the browser-side assertion.
7. **Site integration** — the `/games` card and artwork, the nav entry, the page metadata, and all four online games named on `/privacy-policy` and `/terms` (plus the player-identifier cookie, undescribed since #105).

## The bug the verification found

The plan said to keep `useOrbitAnimation`'s effect "exactly as it is" apart from the comparison. Doing that produces a board that is **wrong online and fine hotseat**, and no HTTP-level assertion can see it.

Online there are two renders per turn carrying the *same* `turnKey`: the optimistic preview, then the server's confirmation of the same turn ~50ms later. React runs the previous effect's cleanup before the next effect body, so the confirmation's cleanup cancelled the orbit the preview had started — and the new value guard, correctly, declined to restart it. The board froze mid-turn, `committed` stopped advancing, and every marble that was never settled kept no id and therefore drew no DOM node.

Measured, on the real page against the real room:

    before:  dom "0C"              server "00 120 141"    (1 marble drawn, 3 held)
    after:   dom "0C 12C 14G"      server "00 120 141"    (agrees, for 12s of resyncs)

The fix is that a sequence's lifetime belongs to the sequence: a new turn cancels the running one explicitly (as it already did), and unmount is handled by `useEffect(() => cancel, [cancel])`. This is now asserted by `park --eval`, so it cannot regress quietly, and `CLAUDE.md`'s claim about that cleanup — which described the old behaviour — is corrected.

## How it was verified

Against a real MySQL 8 (`prisma db push`) and a production `next start`. Docker Desktop was not running; a throwaway container provided the database.

    npm run lint                                   ✔ No ESLint warnings or errors
    npm run build                                  ✓ Compiled successfully
    node .agent/scripts/verify-race-condition.mjs  Race Condition engine: all checks passed
    node .agent/scripts/rc-room.mjs check --presence
      → 35 checks passed, 40 response bodies scanned for tokens

`check` covers, in one run: lobby → playing on the second seat; a 13-turn game with the revision moving by exactly one per accepted action; the winner opening the rematch *asked for by the loser*; an out-of-turn action refused **422** with `"It is not your turn."` as a `string`; a stale revision **409** with the room attached; a third arrival **403** and its snapshot's `me: null`; a mid-game leave that keeps the seat flagged `left` with the board untouched, then a rejoin to the **same slot** with `left: false`; and 45s of silence reading as `connected: false` / `left: false`.

**The refusal assertions were mutation-tested.** Returning the engine's `{ code, message }` object instead of flattening it:

    ok    the refusal is a STRING, not the engine's { code, message } object
    FAIL  and it is the engine's own sentence
          "[object Object]"

Note the type check alone passed — the core stringifies whatever it is handed into the `Error` message, so `typeof message === "string"` is true even when the content is `[object Object]`. Both assertions are needed. The token scan is anchored the same way: it is pointed at a planted payload and must find both tokens before its silence over the run means anything.

Browser states reached, each with a screenshot taken *in* that state (headless Chrome via `shot.mjs`, a fresh profile and therefore a distinct player token per run, playing against `rc-room.mjs park`):

| State | How | Result |
|---|---|---|
| Two players, full game, rematch | `park --seats 1`, browser clicks its own turns | 6 turns from this browser, "Cyan wins" / "Four in a row, after the press.", panel offered Play again, rematch produced a fresh board for both |
| Board closed on the opponent's turn | taps attempted while closed | `tapsThatLandedWhileTheBoardWasClosed: 0`; hint read `Waiting for Cyan…` |
| Spectator | `park --seats 2` | strip: "Watching — this room is full", no seat, no Play again, hint "You are watching this game." |
| Opponent left | `park --seats 1 --leave-after 2` | "Cyan left the game." + Back to the menu |
| Opponent away | `park --seats 1 --drop-after 2` | "Cyan lost connection. They may be back." + "The board picks up where it left off." — no Back to the menu |
| Hazards 1 & 2, on the polling fallback | `EventSource` stubbed to fail before the room mounts, so the hook falls back after three failures | 6 polls over a settled board with `boardReanimatedItself: false`; a staged slide (`4 → 2`) held through 6 more polls, `stagedSurvived12s: true` |
| Hotseat makes no network calls | played 5 turns in "Same device", read resource timing | `apiRoomRequests: []` |

Screenshots: `/games` card, the away strip, the left strip, the spectator board, the post-rematch board. Say the word and I will attach them to the PR; they are not committed.

## Notes for the reviewer

- **One plan detail was wrong and I did not bounce it.** The plan expects a third arrival to be refused `403 room-in-progress`; the core checks capacity first, so a two-seat game answers `403 room-full` and `room-in-progress` is unreachable here (a free seat implies the room is still in the lobby). `useRoom` degrades both to spectating identically, so behaviour matches Connect 404 exactly — only the assertion changed, with the reasoning in the script.
- **`online.opponent` is the cast's name ("Gold"), not the room seat's ("Player 2")**, which is a deviation from the Connect 404 template the plan pointed me at. With a deliberately nameless cast, the core's placeholder would have put "vs Player 2" and "Player 2 left the game." on a screen whose board calls that same person Gold. Flagging it because it is a judgement call, not a copy.
- **The stale-revision notice is my own sentence**, not the core's "The board moved on — try again." — "try again" is the one thing a player must *not* do here, since the indices they chose refer to a board that has since rotated.
- **A browser reload returning to the same seat is asserted at HTTP level, not by literally reloading a tab**: `shot.mjs` mints a fresh profile per run, so two runs are two players. `check` covers the mechanism the reload uses — `pagehide` fires a leave, the rejoin clears `left` and resolves the same slot from the token.
- The spectator screenshot happened to land mid-orbit, so the endgame panel is correctly still held back and the transient winning-line glow is on cells the marbles are passing through. The strip is the part it is evidence for.
- `park` deliberately waits rather than racing the human to the rematch button; simultaneous presses are covered by `check` instead (either seat may ask, and the loser of that race gets no notice).


csalinas-dev:
## 🤖 Review: APPROVED

Checked: security, correctness, efficiency, fit against the plan. Everything below was run against this branch on a real MySQL 8 and a production `next start` — no claim in the PR body was taken on trust, and the two I was most suspicious of were **mutation-tested** to prove they can fail.

**Verified**

*The shared core (the thing that carries the other three games)*
- `git diff origin/epic/race-condition...HEAD -- src/lib/realtime/` is **one file, +2/-0**: the `raceCondition` import and its line in `GAMES`. `git diff --name-only ... 'src/lib/realtime/*'` returns `registry.js` and nothing else. No game rule leaked in — `multiplayer.js` calls `_lib`'s `reducer`/`resetState`/`createState` and restates nothing. `_lib/` and `prisma/` are untouched. Tic Tac Overflow, Edge Case and Connect 404 are unaffected.

*The bug the implementor found — re-verified by reverting it*

This is the one I was told to hit hardest, so I did not accept it from the diff. I restored the plan-as-written cleanup (three `return cancel;`, unmount effect deleted — with anchor assertions so a no-op edit could not read as a pass), rebuilt, and replayed the same browser turn against a real parked opponent:

```
fix reverted:  marblesOnScreen 1   screenDisagreedWithTheRoom 24/24
               onScreen "0C"           inTheRoom "0C 12C 14G"
fix in place:  marblesOnScreen 3   screenDisagreedWithTheRoom 0/24
               onScreen "0C 12C 14G"   inTheRoom "0C 12C 14G"
```

The bug is real, the diagnosis is right, and the fix is what closes it. `check` passed **identically** in both builds, which is exactly the point being made. File restored; tracked tree byte-identical to HEAD afterwards.

*Player identity — and whether the scan can actually fail*
- `rc-room.mjs check --presence` → **35 checks passed, 40 response bodies scanned**, exit 0.
- **Mutation A**: added `token: player.token` to `publicPlayer` in the core, rebuilt. `check` failed on the *first* payload: `FAIL no payload carries a player token — POST /api/rooms (201) exposed $.room.players[0].token, $.room.me.token`, exit 1. The scan fires on a real server-side leak, not just on its own planted anchor.
- **Mutation B**: returned the engine's `{ code, message }` instead of flattening it. Reproduced the warning exactly — `ok the refusal is a STRING …` passed, then `FAIL and it is the engine's own sentence → "[object Object]"`. The `typeof` assertion alone is indeed worthless; the second assertion is what carries it, and it is present.
- No token in any URL: the outbound turn is `{ type, move, place }` with no `slot`; grep for `localStorage` / `document.cookie` / `PLAYER-TOKEN` across the whole game directory returns nothing (identity stays in the core). No `dangerouslySetInnerHTML` or `innerHTML`. `rooms.js` throws `notAPlayer()` before `def.reducer`, so `player.slot` in the `NEW_GAME` branch can never be read off null.

*Room states reached in a real browser, with my own probes (not the PR's)*

| State | Result |
|---|---|
| Full game vs a parked opponent, then rematch | 6 browser turns to a decided ending; `Gold wins` / `Four in a row, after the press.`; Play again → `rematchOpenedByWinner: true`, board cleared in room **and** on screen, no notice |
| Spectator arriving after start | `saysWatching: true`, `hasPlayAgain: false`, `placeableCells: 0`, `me: null`; clicking **all sixteen** cells moved the revision `14 → 14` |
| Deliberate leave | `"Cyan left the game."` + `Back to the menu`, no away wording |
| Silent disconnect | `"Cyan lost connection. They may be back."` + `"The board picks up where it left off."`, no Back-to-the-menu — a genuinely different sentence, both out of `absence.js` |
| Polling fallback (EventSource stubbed to fail *before* the room mounts, via the join form) | `streamAttempts 3`, `pollsIn9s 6` → really on the poll. `boardReanimatedItself: 0`, `screenDisagreedWithTheRoom: 0` over 12s, and a staged slide (cell 0) `stagedSurvived12s: true`. Hazards 1 and 2 both closed |
| Rejoin after refresh | seat `1` → document destroyed (`pagehide` fires the beacon) → new document: **same seat 1**, `left: false`, `"You are G Gold"`, 16 cells, no "left the game" |
| Board closed on the opponent's turn | `tapsLandedWhileClosed: 0` across two runs; hint read `Waiting for Cyan…` |
| Refusal copy on screen | Rewrote the outgoing body in flight to get real verdicts: 409 → `"Your opponent moved first and the board has turned. Take your turn again."`; 422 → `role="alert"` reading **`"There is already a marble there."`**. `objectObjectAnywhere: false` |
| Hotseat | 6 full turns in "Same device" with `fetch` / XHR / `EventSource` / `sendBeacon` all trapped: `allTrappedRequests: []`. Zero network, not just zero `/api/rooms` |

*Site integration*
- `/games` renders six cards; Race Condition is last, same size and construction as the others, `viewBox="0 0 300 300"` matching Connect 404, literal `#4fc1ff` / `#ffd700`, no animation, no external assets. The orbit arrow does the work of distinguishing it from a noughts-and-crosses grid.
- Nav entry present. `/privacy-policy` and `/terms` name all four online games; the cookie correction is factually right — `getPlayerToken` (`useRoom.js:61-86`) writes `localStorage.setItem` *and* `writeCookieToken` on every read.
- `npm run lint` clean · `npm run build` compiled · `verify-race-condition.mjs` all checks passed · PR `MERGEABLE` / `CLEAN`.

**The two flagged judgement calls — both graded correct**

1. **`online.opponent` is the cast name ("Gold"), not the seat name ("Player 2")** — right call, and the safer one. Grepping the whole game directory, `Player 1` / `Player 2` appear only inside comments: no screen renders a core seat name, so there is no contradiction to introduce. The board stamps `G` on every Gold marble, and the browser confirms the strip ("You are ⬤ Gold"), the hint ("Waiting for Cyan…") and the absence sentence ("Cyan left the game.") all agree with it. It also happens to mean no user-suppliable string is rendered at all, since `/join` accepts a `name` this game never asks for.
2. **`403 room-full`, not the plan's `room-in-progress`** — the plan was wrong and the PR is right. `rooms.js:249-250` tests `players.length >= maxPlayers` *before* `status !== "lobby"`, so with two fixed seats `room-in-progress` is unreachable by construction. `useRoom.js:371-376` degrades both to `spectating` in the same branch, so behaviour is identical to Connect 404 — confirmed in the browser above. Correcting the assertion rather than bouncing the plan was the right economy, and the reasoning is in the script.

**Acceptance criteria**
- [x] Two browsers, full game to a decided ending, board closed on the opponent's turn — browser run above
- [x] Rematch by either seat, opened by the winner, no error on a simultaneous press — `check` step 6 + browser
- [x] Third arrival spectates: live board, nothing clickable, no Play again, strip says watching
- [x] Reload returns to the same seat; the other screen does not report a departure
- [x] `left` vs `away` distinguished, both sentences from `absence.js`; no component tests the raw flags
- [x] Illegal action refused with the engine's sentence as a **string** on screen, not `[object Object]` — mutation-tested
- [x] Nothing depends on JSON key order; turn order is the `slots` array
- [x] No re-animation and no wiped draft on the 2s polling fallback
- [x] Hotseat: zero network requests
- [x] `/games` card + nav; `/privacy-policy` and `/terms` name all four games and mention the cookie
- [x] The three #143 carry-overs (`Announcer` consumes `outcome`, `staged` un-shadows the memo, defensive destructuring in `TurnBanner` / `Endgame`)
- [x] lint, build and `verify-race-condition.mjs` clean

**Non-blocking observations**
- `shot.mjs` exits **0** even when a page script returns `ok: false` — the mutated build's `screenDisagreedWithTheRoom: 24` produced exit 0. `park --eval` is therefore a *report*, not a gate, so the `CLAUDE.md` line "asserts the board on screen IS the board in the room" is only true if a human reads the JSON. Worth an exit-code check in `shot.mjs` (or a wrapper) the next time that harness is touched, so this class of regression can fail a script rather than a reading.
- On Windows, `rc-room.mjs`'s `process.exit(1)` mid-`fetch` trips a libuv assert (`!(handle->flags & UV_HANDLE_CLOSING)`) and the process reports **127** rather than 1. Still non-zero so any gate works, but the failure message gets a line of C noise under it.
- During the reload window the seat really is `left: true` for about 1.5s, so the other player can see "Gold left the game." flash before the rejoin clears it. That is the core's `pagehide` beacon, shared with all three existing games and not this PR's to fix — noting it only so it is not mistaken for a Race Condition bug later.
- `/privacy-policy`: "keeping both is what stops a refresh in a private window turning you into a different player" is a slightly off reading of the code's own reason, which is that either store may be *blocked* (`useRoom.js:65-80`) — private browsing is one instance of that rather than the mechanism. One clause, and only if the page is edited again.
- `notice` persists until the next accepted action, so a stale-revision message can outlive the situation that produced it by a turn. Matches the existing games; mentioned for completeness.

Nothing blocking. This is the rare large PR where the risky part is small and was actually driven, and the one bug in it was found by the author's own instrumentation rather than by a reviewer — which is the outcome the verification section was written to produce.

PR #148, opened 2026-08-16
epic(games): Race Condition — Orbito, hotseat and online

### LINKED ISSUE #141: feat(race condition): build game

csalinas-dev:
Integration PR for epic #141. Merges `epic/race-condition` into `main`.

Closes #141.

### What this is

**Race Condition** — an Orbito variant at `/games/race-condition`, playable on one device or in a room with somebody else, in the same visual language as Connect 404.

Orbito's turn is: optionally slide one of your **opponent's** marbles to an adjacent empty square, place one of your own, then press the button and every marble on the board orbits one step. A line only counts *after* the press — so you can hand your opponent the win by rotating their four into place.

### What is in this diff, and what is not

The epic was built in three parts. **Part 1 (the rules engine, #142) is not in this diff** — it merged directly into `main` before the epic-branch rule existed, and `epic/race-condition` was cut from `main` afterwards. Nothing is missing from the feature; the engine is simply already on `main`. This PR is parts 2 and 3:

| | | |
|---|---|---|
| #142 | rules engine, pure `_lib/` | already on `main` (PR #145) |
| #143 | animated board, hotseat play | PR #146 → this branch |
| #144 | online multiplayer, `/games` card | PR #147 → this branch |

### Review history

Each part was reviewed independently before merging here. Worth carrying into this review rather than repeating:

- **#142** — the reviewer re-extracted the publisher's rulebook PDF itself, re-derived the orbit permutation from scratch, and ran an independently written shadow engine lock-step against the real one for 4000 games, then flipped the real engine to clockwise to prove the cross-check could diverge.
- **#143** — one blocking bug found and fixed: the turn banner named the winner from `game.turn`, which is the player who *ended* the game, not the winner when the orbit completes the opponent's line. 38.5% of wins hit it. Fixed by collapsing four derivations into one.
- **#144** — the plan's own hazard fix, implemented as written, still broke the board online (two renders per turn share a turn key; React ran the effect cleanup and killed the orbit at frame 0). Found only in a browser against a live room; every HTTP assertion passed throughout. The reviewer reverted the fix to confirm it: 24/24 board disagreements without it, 0/24 with it.

### Verification available to this review

- `node .agent/scripts/verify-race-condition.mjs` — engine self-play, and it is mutation-tested
- `node .agent/scripts/rc-ending.mjs` — seeded games that reach specific endings (orbited-in win, simultaneous draw, exhausted draw) and replay them by clicking the real board
- `node .agent/scripts/rc-room.mjs check|park` — asserts the HTTP contract including that no payload carries a `token`, and plays the opponent's side so a real browser can be the other player

### Known, and deliberately not fixed here

`shot.mjs` exits 0 even when a page script reports `ok: false`, so `park --eval` is a report rather than a gate. Filed separately.


csalinas-dev:

Epic-level review of the merged branch, not a re-run of the three part reviews. Checked out `epic/race-condition` at 702cde2 (fast-forward of `main`, no merge commit, `MERGEABLE`), built it, and played it — hotseat and online, against a real second player.

One blocking item, and it is exactly the class of thing this review exists for: **a break that neither part's review could have seen, because it is the seam between them.**

### Blocking

1. **`.agent/scripts/rc-ending.mjs:33` (the generated page script) — the mode picker from #144 stops it reaching the board, and it fails with exit code 0**

   At #146's head, `/games/race-condition` rendered the hotseat board directly (`RaceCondition.jsx` was `<ContextProvider><Game/></ContextProvider>`), so a page script could start clicking cells. #147 put `Lobby.jsx` — "Same device" / "Play online" / "I have a code" — in front of it and did not update the script, which still opens with:

   ```js
   if (cells().length !== 16) return { error: "board not found" };
   ```

   The tool's own documented invocation, printed by the script itself:

   ```
   $ node .agent/scripts/rc-ending.mjs orbited-in --eval OUT.js
   $ node .agent/scripts/shot.mjs http://localhost:3000/games/race-condition 900 out.png OUT.js
   {"viewport":{...},"result":{"error":"board not found"}}
   $ echo $?
   0
   ```

   Zero turns are played and nothing about the ending is inspected. Combined with the `shot.mjs` exit-code problem this PR's own body flags, the result is a check that **reports nothing and looks green** — and `CLAUDE.md` (added in this diff) tells the next agent "Reach for it before believing any claim about how a finished game renders." The one blocking bug in #143 was a rendered-ending bug found only because this script existed; on `main` it will no longer find one.

   Fix: wait for the picker to hydrate and click through it before the bail-out. I verified this exact prelude in `pageScript`:

   ```js
   const hydrated = (el) => el && Object.keys(el).some((k) => k.startsWith("__reactProps"));
   if (cells().length !== 16) {
     let btn = null;
     for (let i = 0; i < 100; i += 1) {
       btn = [...document.querySelectorAll("button")].find((b) => /Same device/.test(b.textContent));
       if (hydrated(btn)) break;
       await sleep(100);
     }
     if (!hydrated(btn)) return { error: "mode picker never hydrated" };
     btn.click();
     for (let i = 0; i < 60 && cells().length !== 16; i += 1) await sleep(50);
     await sleep(300);
   }
   if (cells().length !== 16) return { error: "board not found after picking same device" };
   ```

   Clicking before hydration silently does nothing (I hit that first), so the `__reactProps` wait is load-bearing, not defensive padding. With it, all three endings replay through the real board:

   | ending | turns | banner | endgame panel |
   |---|---|---|---|
   | orbited-in | 12 | `Cyan wins` / "Gold turned the board into Cyan's line." | "Cyan opens the next game." |
   | simultaneous | 15 | `Draw` / "Both lines completed on the same press." | "A tie is a tie." |
   | exhausted | 16 | `Draw` / "The board filled and nothing was found." | "…five more presses found nothing." |

   Please also make the script's failure loud (non-zero, or an `ok: false` the caller must read) so the next silent no-op is not silent.

### What I verified working — no need to re-check any of this

**Hotseat, end to end.** Full 12-turn `orbited-in` game clicked through the real board in headless Chrome. The banner named **Cyan** (the winner) while **Gold** ended the game, ringed cells `[0,5,10,15]`, marbles `0C 5C 10C 15C` with the diagonal highlighted, announcer correct. No console errors or warnings captured across a whole game. Overtime (6 presses) plays every press.

**Online, end to end.** Against `rc-room.mjs park` holding seat 0: the browser took seat 1, played 6 turns, the room reached 13 marbles, **Gold won on Cyan's press**, "Gold opens the next game", then `Play again` → new board, revision 16, turn = the winner. Separately, `park --eval`: `screenDisagreedWithTheRoom: 0`, `boardMovedOnItsOwn: 0` over 24 samples, `ok: true` — the `turnKey` / no-cleanup fix from #144 holds on the merged branch.

**Scripts.** `verify-race-condition.mjs` all checks pass (2000 self-played games). `rc-room.mjs check` 34/34, 22 bodies scanned for tokens. `rc-ending.mjs` itself (engine half) produces all three endings.

**Build and lint.** `npx next lint` clean. `npm run build` succeeds; `/games/race-condition` prerenders static at 13.6 kB.

**Blast radius.** Outside `games/race-condition/` this branch touches only `registry.js` (+2), `Nav.jsx` (+1), `games.js` (+7), the two policy pages, `CLAUDE.md`, and the scripts. `multiplayer.js` pulls in only `_lib` and `context/actions.js`, both free of React, so nothing new lands in the API route bundle. The other three games still work on this branch: TTO placed a mark, Connect 404 dropped a piece, Edge Case rendered its 6×6 setup; all four games create rooms (201), and a browser joining a room resolves its seat correctly (`You are Blue vs Player 1` on Connect 404, `Player 2 you` on Edge Case).

**Site integration.** Searching for every place a game is listed outside its own directory returns exactly `games.js`, `Nav.jsx`, `registry.js` — all three updated. No sitemap/robots to update. Privacy and terms now name all four online games (the #107 precedent), and the added cookie sentence is accurate — I confirmed the browser holds `CSALINAS-PLAYER-TOKEN` in both local storage and a cookie. The `/games` card renders in the row, its line is a real winning line (cells 1/5/9/13, a column), and its arrow points left along the top edge — counter-clockwise, matching `ORBIT_TO = [4,0,1,2,8,9,5,3,12,10,6,7,13,14,15,11]` (0→4 down the left, 1→0 left along the top).

**Coherence with the engine already on `main`.** The only `_lib` change is `applyMove` extracted out of `playTurn` and re-exported from `_lib/index.js`, with `playTurn`'s `played === cells` reference identity preserved — `move` still defaults to `null`, so `applyMove` sees `null` and returns `cells` by reference exactly as before. No duplicated rule, no stale import, and the added `CLAUDE.md` sections match what the code does.

### Non-blocking

- **A pre-existing core bug worth its own issue (NOT this PR's).** On a cold dev server (`rm -rf .next`, first request), a player who legitimately holds a seat can land in a state where the strip says "You are ? grey vs your opponent", the banner says "Waiting for your opponent…", and the board never opens — while the server says `me.slot === 1` and the seat is real. The cause is in `src/lib/realtime/`: `redeemTicket` returns `null` for a ticket it cannot honour and `stream/route.js` then serves an **anonymous** stream (`token = undefined`), whose `sync` payloads carry `me: null`; `useRoom.applyRoom` overwrites the seat with them and never recovers, because the stream counts as connected so the polling fallback (which does send `X-Player-Token`) never starts. Presence confirms it — the seat shows `connected: false` while payloads keep arriving. **I reproduced the identical failure on `/games/connect-404`, which is on `main` and untouched by this branch**, so it is not a Race Condition regression; warm, both games are fine across 5/5 runs. Suggested direction: have the stream route refuse (401) a ticket that was presented and could not be redeemed, rather than silently downgrading to spectator.
- `CLAUDE.md`'s realtime-core bullet still says identity lives "in `localStorage` under `CSALINAS-PLAYER-TOKEN`". This PR correctly fixes exactly that staleness in the privacy policy (cookie *and* local storage since #105) and leaves it in the doc a few lines above. One sentence while you are in the file.

csalinas-dev:
Addressed, in 836f907. Scoped to the script and the doc — **no game code is touched**, so your hotseat/online/blast-radius verification stands unchanged.

### Blocking — `rc-ending.mjs`'s generated page script

**Fixed, and re-verified from scratch rather than assumed.** I took your prelude as the design but re-ran all three endings in my own tree rather than trusting that it transfers, since that assumption is the same class of mistake as the bug.

The prelude now waits for the picker to hydrate, clicks **Same device**, waits for the 16 cells, and only then bails:

```js
const hydrated = (el) =>
  !!el && Object.keys(el).some((key) => key.startsWith("__reactProps"));
```

Confirmed your point that the hydration wait is load-bearing: the page is prerendered, so the button exists in the markup well before React owns it, and a click on it does nothing at all — no handler, no error, no board.

**All three endings replay through the real board** (`npx next start`, production build, port 3141), and they match your table exactly:

| ending | `ok` | turns | banner | endgame panel | ring |
|---|---|---|---|---|---|
| orbited-in | `true` | 12 | `Cyan wins` / "Gold turned the board into Cyan's line." | "Cyan opens the next game." | `[0,5,10,15]` |
| simultaneous | `true` | 15 | `Draw` / "Both lines completed on the same press." | "A tie is a tie." | `[0,2,4,6,8,10,12,14]` |
| exhausted | `true` | 16 | `Draw` / "The board filled and nothing was found." | "…five more presses found nothing." | `[]`, 16 marbles |

Each ring is the engine's own line: `orbited-in`'s `[0,5,10,15]` is the `r1c1-r2c2-r3c3-r4c4` diagonal it searched for, and `simultaneous`'s eight cells are both columns (`0,4,8,12` and `2,6,10,14`). `orbited-in` still ends with **Gold** and is won by **Cyan**.

### Blocking — "make the failure loud"

**Done, within the script.** Every return now carries `ok`, and it is false unless the replay reached the board, played every turn, **and** arrived at a finished game:

```js
ok: played === TURNS.length && finished,
```

The `finished` anchor (the endgame panel is on screen) is there for the failure mode your finding is really about: if the clicks land on a board that is not listening, the turn loop still runs to the end and the report still comes back full of plausible-looking nulls. That has to read as a failure, not as a result.

**Negative control** — the same generated script pointed at a page with no picker and no board:

```
$ node .agent/scripts/shot.mjs http://localhost:3141/ 900 - rc-orbited-in.js
{"viewport":{...},"result":{"ok":false,"error":"the mode picker never hydrated"}}
$ echo $?
0
```

A green run can now fail, which is the part the old script could not do.

I did **not** make it non-zero, because that is `shot.mjs`'s exit code and it is #150's — fixing it here would edit a script shared by every game on a Race Condition PR. Instead the obligation to read `ok` is now stated in the two places a caller looks: the script's own header (`READ result.ok FROM THAT SECOND COMMAND`) and the `CLAUDE.md` bullet. When #150 lands, `ok: false` is exactly the signal `shot.mjs` needs to exit non-zero on.

### Non-blocking

- **`CLAUDE.md` identity staleness** — fixed. The realtime-core bullet now says the token lives under `CSALINAS-PLAYER-TOKEN` in **both** `localStorage` and a cookie of the same name, with the reason (a refresh in a private window would otherwise make you a different player), matching the privacy policy wording in this same diff.
- **`CLAUDE.md`'s `rc-ending.mjs` bullet** — updated beyond the one sentence, since it was the instruction that would have sent the next agent into the silent no-op. It now records that the script drives the mode picker, that `result.ok` and not the exit code is the thing to read, and that this script has already been silently green once.
- **The cold-start `redeemTicket` core bug** — agreed it is out of scope here, and your reproduction on `/games/connect-404` (untouched by this branch) is what settles it. Not touched; it wants its own issue against `src/lib/realtime/`.

### Verification

    node .agent/scripts/rc-ending.mjs {orbited-in,simultaneous,exhausted} --eval …   engine half: 12 / 15 / 16 turns, unchanged
    npm run build                                                                    clean; /games/race-condition static, 13.6 kB
    npx next lint                                                                    No ESLint warnings or errors
    node .agent/scripts/shot.mjs …/games/race-condition 900 out.png <each eval>      ok: true on all three
    node .agent/scripts/shot.mjs http://localhost:3141/ 900 - <orbited-in eval>      ok: false, "the mode picker never hydrated"

csalinas-dev:
## 🤖 Review: APPROVED

Re-review of `836f907` only. Scoped to the fix, as directed — hotseat/online play, the blast radius outside the game directory, and the other three multiplayer games were verified on the previous pass and are not re-litigated here.

**Verified**

Fresh detached worktree at `836f907`, `npm ci`, `prisma generate`, `npm run build` (clean), `npx next start -p 3141`. Endings regenerated from scratch with `rc-ending.mjs --eval`, not taken from the PR description.

*The three endings drive through the picker to the board:*

| ending | turns | `ok` | outcome read off the DOM |
|---|---|---|---|
| `orbited-in` | 12 | `true` | "Cyan wins" / "Gold turned the board into Cyan's line." — ringed `[0,5,10,15]`, and the announcer says "on Gold's press", i.e. the winner is not the last mover. This is the exact case the banner bug shipped on. |
| `simultaneous` | 15 | `true` | "Draw" / "Both lines completed on the same press." — ringed `[0,2,4,6,8,10,12,14]`, both columns, neutral `—` chip with a transparent background. |
| `exhausted` | 16 | `true` | "Draw" / "The board filled and nothing was found." — 16 marbles, `ringed: []`, announcer reports 6 presses. |

Matches the implementor's table exactly, independently reproduced.

*The check can fail — three distinct paths, each made to fire:*

1. `http://localhost:3141/privacy-policy` → `{"ok":false,"error":"the mode picker never hydrated"}`. As claimed.
2. `http://localhost:3141/games/tic-tac-overflow` → `{"ok":false,"error":"board not found after picking Same device"}`. TTO's picker *does* hydrate and *does* yield a board, so this also confirms the `__reactProps` wait resolves positively on a real hydrated picker rather than being vacuously false — the same thing the three green runs prove from the other direction.
3. The endgame anchor, which is the load-bearing claim. I served an inert page carrying 16 `[data-cell]` buttons with real placement labels, a visible banner and a hidden announcer, and nothing listening. The turn loop ran to completion and the report came back exactly as advertised:

```json
{ "ok": false,
  "error": "the game never finished — the clicks did not reach the board",
  "turnsPlayed": 12,
  "banner": { "headline": "Cyan's turn", "hint": "Place a marble.", "chip": "C" },
  "endgame": null, "announcer": "nothing has happened",
  "ringed": [], "marbles": [] }
```

`turnsPlayed: 12` of 12 and a banner that reads like a game — this is the "plausible nulls" pass the anchor exists to convert into a failure, and it does. `shot.mjs` exited **0** on all three negative controls, which is precisely why `ok` had to become the contract.

*The `CLAUDE.md` edits:*

- The `rc-ending.mjs` bullet now carries **"Read `result.ok`, not the exit code"** with the #150 reason and the record of the silent green. Correct.
- The realtime-core identity sentence checked against `src/lib/realtime/useRoom.js`: `getPlayerToken()` reads `localStorage` then `readCookieToken()`, re-asserts to **both** on every read so a half-available browser heals, and `TOKEN_COOKIE_MAX_AGE` is a year. The new wording — "in **both** `localStorage` and a cookie of the same name (either alone is enough; keeping both is what stops a refresh in a private window turning you into a different player)" — matches the code and its #105 rationale line for line, and matches `privacy-policy/page.js:68` as amended in `702cde2`. The #149 discrepancy is closed on this file.

*Blast radius of the fix:* `git diff --stat 836f907^ 836f907` is `.agent/scripts/rc-ending.mjs | 70 ++++-` and `CLAUDE.md | 4 +-`, nothing else. No game code. `git diff --name-only 836f907^ 836f907 -- .agent/scripts/shot.mjs` is empty — `shot.mjs` was left alone for #150, which is the right call.

**Acceptance criteria**
- [x] `rc-ending.mjs` reaches the board through the mode picker — all three endings replayed to a finished game on a clean build.
- [x] The check can fail — three fail paths reproduced, each with the right sentence.
- [x] `ok` asserts the endgame panel, not just the turn count — proven against a board that swallows every click.
- [x] `CLAUDE.md` tells the next agent to read `result.ok`, and its identity sentence matches the code.
- [x] No game code, no `shot.mjs` edit.

**Non-blocking observations**

- The `played === TURNS.length` half of `ok` is unreachable as written: the turn loop has no `break` and increments `played` unconditionally, so it always equals `TURNS.length` if the loop returns at all. A missing `cellAt()` throws instead, which surfaces as a `shot.mjs` exit 1 — loud, so nothing is lost. `finished` is doing all the work, and it works. Worth knowing only so nobody later reads that ternary as live coverage.
- Waiting out the full hydration loop costs ~10s before `"the mode picker never hydrated"` returns. Fine for a script you run deliberately.

*(Posted as a comment rather than an approving review: GitHub will not let the PR author's account approve its own PR.)*

PR #157, opened 2026-08-16
fix(realtime): refuse an unredeemable stream ticket instead of serving it anonymously

### LINKED ISSUE #155: fix(realtime): a rejected stream ticket silently serves an anonymous stream and locks the player out of their seat
Found by the reviewer of epic #141 while reviewing PR #148. **This is not that PR's regression — it was reproduced identically on `/games/connect-404`, which is on `main` and untouched by that work.** It affects every multiplayer game on the site.

### What happens

When `redeemTicket` cannot honour a ticket, `src/lib/realtime/stream/route.js` does not refuse the connection — it silently serves an **anonymous** stream. The consequences cascade:

1. `sync` payloads then carry `me: null`.
2. `useRoom.applyRoom` overwrites the player's own seat with that null.
3. The board never opens again. The player sees "You are ? grey" and "Waiting for your opponent…" while sitting in a room they are legitimately seated in.
4. **Polling never rescues them**, because the anonymous stream counts as connected — so the fallback that exists for exactly this situation never engages.

The player's only recovery is a reload, and nothing on screen suggests that.

### Why it deserves priority over cosmetic work

It is a live defect in shipped games (Tic Tac Overflow, Edge Case, Connect 404), it presents as "the game is broken" rather than as an error, and the failure is *silent by construction* — an unhonoured ticket is treated as a valid anonymous viewer rather than as a failure. Tickets are single-use and time-limited, so this is reachable in ordinary use: a retried connection, a slow network, a reconnect after sleep.

### What the fix has to settle

- A ticket that cannot be redeemed should fail the stream, not downgrade it. Decide the status and what the client does with it.
- `applyRoom` overwriting a known seat with `me: null` is a second, independent bug — a payload that does not know who you are should not be able to un-seat you. Fixing only the route leaves that landmine.
- Whatever is done, the client must end up in a state where polling *can* rescue it.

### Done when
- An unredeemable ticket produces an explicit failure rather than an anonymous stream.
- A client that receives a payload without an identity does not lose the seat it already holds.
- The scenario is demonstrated failing before the fix and passing after — `.agent/scripts/rc-room.mjs park` can hold the opponent's side while the case is forced.
- Confirmed on more than one game, since the bug is in shared code.

---

## 🤖 Implementation Plan

_Written by the architect agent on 2026-08-16 01:34 UTC. The implementor builds exactly this._

### Goal

A stream credential that cannot be honoured fails the connection with an explicit
**403 `bad-ticket`** instead of quietly serving an anonymous stream, and a room payload
that does not know who the caller is can no longer take away a seat the client is
already sitting in. After this ships, the worst an unredeemable ticket can do is cost a
player a few seconds of "Reconnecting…" before the existing polling fallback picks them
up — still seated, still able to move.

Both halves are in scope. They fix different things and neither covers the other: the
route fix is the only thing that restores *presence* (an anonymous stream never calls
`markPresent` for the player, so everyone else counts them out after 45s), and the
client fix is the only thing that protects the seat from *any* future identity-less
payload. The issue is right that the client half is the more dangerous of the two.

---

### The bug, reproduced

Verified on `origin/main` (`0b02794`) against a real dev server **and against a real
production build** (`next build && next start`), with a real MySQL, using a throwaway
harness that mints a ticket and then opens the stream four ways. Both runs printed the
same thing:

```
create: 201 9CSA
join B: 200 status = playing
ticket: 200 issued

[1] fresh ticket   -> 200 text/event-stream | me: slot 0     <- anchor: the probe works
[2] REPLAYED ticket-> 200 text/event-stream | me: null       <- THE BUG
[3] bogus ticket   -> 200 text/event-stream | me: null       <- THE BUG
[4] no ticket      -> 200 text/event-stream | me: null       <- correct (spectator)

snapshot with A's token -> 200 me: 0 players: [[0,true],[1,true]]   <- A is really seated
```

`[2]` and `[3]` are the defect: the server hands a *seated player* a stream that reports
`me: null`, with a 200 and a `text/event-stream` content type — indistinguishable from a
spectator connection.

Second finding, worth knowing because it is probably how the PR #148 reviewer hit it in
ordinary use: in `next dev`, the **first** stream open after the stream route is compiled
also fails to redeem — the module instance holding `tickets.js`'s `Map` is rebuilt during
that compilation, so the ticket the (already-compiled) ticket route minted is not in the
Map the stream route reads. Run [1] once against a cold dev server and it reports
`me: null`; run it again and it reports `me: slot 0`. It does **not** happen in a
production build (case [1] passed on the first request after `next start`). Same class of
failure as a deploy restart, just far easier to trigger.

The cascade after that, all confirmed by reading the code:

1. `src/app/api/rooms/[code]/stream/route.js:40` — `redeemTicket(ticket, code) || undefined`
   turns a refusal into "anonymous", and every `roomPayload(row, undefined)` that stream
   emits carries `me: null` (`src/lib/realtime/serialize.js:32`).
2. `useRoom`'s `applyRoom` (`src/lib/realtime/useRoom.js:167`) accepts it wholesale, so
   `room.me` goes null on a client that got a perfectly good `me` from its own join.
3. In Connect 404 `me?.slot ?? null` → `youSlot = null`
   (`src/app/(pages)/games/connect-404/online/OnlineGame.jsx:49`), so `Game.jsx:62`
   computes `interactive = false` and the board is dead; `OnlineBar` renders
   `playerFor(players, null)`, whose fallback descriptor is literally
   `{ initial: "?", colorName: "grey" }` (`connect-404/players.js:87`) — the "You are ?
   grey" in the report — and `opponent` falls back to the string "your opponent".
   `send()` also short-circuits on `if (!current.me) return { error: "not-a-player" }`
   (`useRoom.js:437`), so even the keyboard path is dead.
4. Nothing rescues it: the stream opened with a 200, so `es.onerror` never fires,
   `connected` stays true, `failures` stays 0 and the polling fallback — which sends
   `X-Player-Token` and would have restored `me` on its first tick — never engages.
5. And on the *server*, `subscribeRoom(code, undefined)` makes `markPresent` a no-op
   (`presence.js:23` returns early on a falsy token), so the other players watch this
   one go "away" after `PRESENCE_STALE_MS`.

Tic Tac Overflow has the same shape (`me` → `mark`/`opponent`), and Edge Case already
half-knew about this — `edge-case/online/OnlineGame.jsx:73-83` carries a comment that
names "a stream whose ticket did not redeem" as the reason `me: null` cannot be trusted,
and works around it with a `heldSlot` ref *for the wording only*. That workaround is the
prior art this plan generalises into the core so all three games get it.

---

### Approach

**1. The stream route refuses a ticket it cannot redeem.**

A `ticket` query parameter that is present but does not redeem is a failure, not a
downgrade. An *absent* ticket stays exactly what it is today — the spectator path — so
case [4] above must keep working unchanged.

Status: **403 with `{ "error": "bad-ticket", "message": ... }`**, JSON, no stream.
Chosen over 401 to stay inside the vocabulary this repo already documents (403 is
already `not-a-player` / `room-full` / `room-in-progress` — "the credential you showed
does not entitle you to this"), and because 401 obliges a `WWW-Authenticate` header this
API has no scheme for. It is worth being clear that **`EventSource` cannot read the
status code**: any non-200 fails the connection, fires `error`, and — unlike a dropped
connection — does *not* auto-reconnect. So the status is a contract for humans, curl and
the verification harness; the client behaviour it produces is the generic error path.

That path already recovers, which is the point:
`es.onerror` → `failures += 1` → after `retryDelay(failures)` it calls `openStream()`
again, which **mints a brand-new ticket** (`useRoom.js:270`). So the ordinary
legitimate failures — a ticket that expired in a backgrounded tab, one consumed twice, a
server that restarted and lost the Map, a dev recompile — all self-heal on the first
retry (~1s). A redemption that fails *every* time burns the three-failure budget in ~3s
and lands on `startPolling()`, which reads `GET /api/rooms/[code]` with
`X-Player-Token`, restores `me`, marks the player present again, and is playable at 2s.
No case becomes terminal. Do not add a special retry schedule for this — the existing
one is already correct for it.

**2. `applyRoom` stops letting an identity-less payload un-seat you.**

The rule, generalised from the Edge Case comment: *the roster cannot lie the way `me`
can.* If a payload says `me: null` but its own `players` array still contains the slot
this browser was last told it holds, the payload simply did not know who was asking —
keep the seat, taking the seat object out of `next.players` so name/colour/`connected`/
`left` all stay fresh. If the slot is **gone from `players`**, the seat really is gone
(host removed them, or they left in the lobby — the only two ways a seat disappears) and
`me: null` is the truth and must be rendered.

This is deliberately *not* "ignore payloads without an identity": the state in such a
payload is still authoritative and a game must not stop updating. Only `me` is repaired.

Rejected alternatives, so nobody re-litigates them in review:

- *Authenticate the stream from the `CSALINAS-PLAYER-TOKEN` cookie instead of a ticket.*
  It would work (the cookie exists since #105 and `EventSource` sends it same-origin),
  but it makes the ticket decorative and rewrites a documented security contract to fix
  a defect that does not need it. Out of scope. In particular **do not add a
  cookie fallback when a ticket fails** — that recreates "silently identifies you
  through a path the contract says is not used" one layer down.
- *Make tickets multi-use or long-lived.* Directly undoes why they exist.
- *Have the client detect `me: null` on a stream and treat it as a transport failure.*
  Redundant once the route refuses, and it would misfire on genuine spectators.

---

- `src/lib/realtime/README.md` is the contract for this directory; it must end up
  describing whatever you build (status table, stream row, and the identity rule).
- **No game rules under `src/lib/realtime/`** — this change adds none, but the new
  helper must stay pure and dependency-free (`constants.js` carries the same warning:
  `useRoom` imports from it, so anything it imports reaches the browser bundle).
- `src/lib/realtime/tickets.js:62` `redeemTicket(ticket, code)` returns the token or
  `null`; it deletes the entry before validating, which is what makes replay fail.
- `src/lib/realtime/errors.js` is where every failure the room layer can produce is
  defined as a `RoomError` factory, and `toResponse(err)` is the only thing routes call.
- `src/lib/realtime/useRoom.js:167` `applyRoom` is the single funnel every payload
  passes through — SSE `sync`, the polling tick, `send`'s 409/422 body, `leave`,
  `refresh`. Fixing it here fixes all of them.
- Prior art to imitate and then simplify: `src/app/(pages)/games/edge-case/online/OnlineGame.jsx:59-83`.
- This repo has **no test runner** (`package.json` has only dev/build/start/lint).
  Verification is a committed script under `.agent/scripts/`, following
  `verify-race-condition.mjs`: one line per check, an anchor assertion per case, exit
  non-zero on any failure. `.agent/scripts/lib/esm-resolver.mjs` is the resolve hook that
  lets bare `node` import this repo's extensionless relative imports (import the module
  under test **dynamically** after `register(...)`).

**Getting a database in a worktree** (the `.env` provisioned into an agent worktree does
not carry a usable `DATABASE_URL`; this is what I used, and it does not touch anybody
else's data):

```bash
docker run -d --name issue155-mysql -e MYSQL_ROOT_PASSWORD=issue155 \
  -e MYSQL_DATABASE=csalinas -p 127.0.0.1:13155:3306 mysql:8.0
# wait for: docker exec issue155-mysql mysqladmin ping -h 127.0.0.1 -uroot -pissue155
export DB='mysql://root:issue155@127.0.0.1:13155/csalinas'
npm ci && npx prisma generate
DATABASE_URL="$DB" npx prisma db push --skip-generate
DATABASE_URL="$DB" ./node_modules/.bin/next dev -p 3155
docker rm -f issue155-mysql        # when you are done
```

An explicit `DATABASE_URL` in the environment wins over the one in `.env` — Next does
not override variables that are already set.

Reading SSE from bare Node (no `EventSource` needed, and it lets you see the status and
content type, which `EventSource` hides):

```js
const ctrl = new AbortController();
const res = await fetch(url, { signal: ctrl.signal });   // res.status, res.headers
const reader = res.body.getReader(); const dec = new TextDecoder();
let buf = ""; while (buf.indexOf("\n\n") === -1) {
  const { value, done } = await reader.read(); if (done) break;
  buf += dec.decode(value, { stream: true });
}
ctrl.abort();   // one event is all we need
```

---

### Changes

1. **`src/lib/realtime/errors.js`** — add a factory beside the other 403s:
   `badTicket()` → `new RoomError("bad-ticket", "This stream link is no longer valid — reconnect to get a new one.", 403)`.
   Comment it with *why* it is not a downgrade to anonymous.

2. **`src/app/api/rooms/[code]/stream/route.js`** — replace
   `const token = redeemTicket(ticket, code) || undefined;` (line 40) with: if the
   `ticket` param is absent or an empty string, `token` stays `undefined` (spectator,
   unchanged); otherwise redeem it and, when redemption yields nothing,
   `return toResponse(badTicket())` before any stream is constructed. Update the
   handler's doc comment — it currently says the ticket "only decides who `me` is",
   which is exactly the assumption that made this a silent failure.

3. **`src/app/api/rooms/[code]/ticket/route.js`** — require a token. Today a body with
   no `token` still mints a ticket, and after change 2 that ticket would redeem to
   `undefined` and 403 — a confusing dead end nobody can act on. Reject it up front with
   the existing `bad-token` 400 vocabulary. No in-repo caller does this (`useRoom`
   only mints when it has a token), so this closes a hole rather than changing behaviour.

4. **`src/lib/realtime/identity.js`** (new, pure, no imports) — export
   `keepSeat(payload, heldSlot)`:
   returns `payload` unchanged when it is falsy, when `payload.me` is set, when
   `heldSlot` is null/undefined, or when no entry in `payload.players` has that slot;
   otherwise returns `{ ...payload, me: <that entry from payload.players> }`. Header
   comment carries the rule in the words of the Approach section above — this file is
   the one place that decides when `me: null` is believable, the way `absence.js` is the
   one place that decides how absence is worded.

5. **`src/lib/realtime/useRoom.js`** — add a `seatRef` (`useRef(null)`) and route every
   payload through `keepSeat` inside `applyRoom`: after the monotonic-revision guard,
   `const merged = keepSeat(next, seatRef.current);` then `seatRef.current = merged.me?.slot ?? null;`
   and store/`setRoom` the merged payload. Note in a comment that clearing the ref when
   the merged payload really has no seat is what makes a lobby "leave"/host-removal
   still land: the seat is deleted from `players`, so `keepSeat` declines to restore it.
   No other call site changes — `send`, `refresh`, `leave` and both transports already
   go through `applyRoom`.

6. **`src/app/(pages)/games/edge-case/online/OnlineGame.jsx`** — comment only, at
   lines 73-78: the "a stream whose ticket did not redeem" case is now handled in the
   core, so say that and say that the `heldSlot`/`removed` check stays as the thing that
   tells *being removed* apart from *standing up*. **Do not delete the `removed` logic** —
   it answers a different question and is still load-bearing.

7. **`.agent/scripts/verify-realtime-stream.mjs`** (new) — the harness. Two sections,
   one line per check, exit non-zero on any failure:
   - *Pure*: `keepSeat` — restores `me` when the held slot is still in `players`;
     leaves `me: null` when the slot is gone; never overwrites a non-null `me`;
     no-ops on `heldSlot === null`; returns a *copy* rather than mutating the payload.
     Import it dynamically through `lib/esm-resolver.mjs`.
   - *Live*, against `--base` (default `http://127.0.0.1:3000`): create a room, join a
     second player, mint a ticket, then the four cases from the reproduction above —
     fresh ticket **200 + `me.slot === 0`** (this is the anchor; if it fails, say so and
     stop rather than reporting the rest), replayed ticket **403 `bad-ticket`**, unknown
     ticket **403 `bad-ticket`**, no ticket at all **200 SSE with `me: null`**. Plus
     `POST /ticket` with `{}` → 400. If the base URL is unreachable, **fail** — do not
     "skip" the live half, a skip reads as a pass.
   Header comment: what it does, how to run it, and the docker/prisma recipe above.

8. **Docs** — `src/lib/realtime/README.md`: add `403 bad-ticket` to the status table;
   amend the stream row and the ticket paragraph to say a present-but-unredeemable
   ticket is refused while an absent one is still the spectator path; add a short
   "Identity in a payload" note stating the `keepSeat` rule and that `me: null` is only
   believable when the seat is also gone from `players`. `CLAUDE.md`: one bullet in the
   realtime run of bullets for the same rule, and one bullet for the new script (the
   file already documents `verify-race-condition.mjs` and `shot.mjs` this way).

---

### Out of scope

- Cookie-authenticated streams; any change to how tickets are minted, stored, or how
  long they live; moving them out of memory.
- `SSE_FAILURES_BEFORE_POLLING`, the backoff schedule, or anything else about the
  transport's retry policy.
- Presence (`presence.js`, `absence.js`) — this change fixes presence for the affected
  player by making the stream carry a token again; it does not touch the mechanism.
- Refactoring Edge Case's `removed` / `heldSlot` logic beyond its comment, and any other
  game-side change. If a game needs a code change to benefit from this, the core fix is
  in the wrong place.
- The dev-only "first stream after a recompile loses the Map" amplifier. Record it in
  the PR description; it is a `next dev` module-instantiation artifact (it does not
  happen in a production build), and after this change it produces a loud 403 and a
  one-retry recovery like every other lost ticket.

---

### Acceptance criteria

- [ ] `GET /api/rooms/<code>/stream?ticket=<already-redeemed>` answers **403** with a
      JSON body `{"error":"bad-ticket", ...}` and `content-type: application/json`; no
      SSE stream is opened.
- [ ] Same for a ticket that never existed (expired / lost on restart).
- [ ] `GET /api/rooms/<code>/stream` with **no** `ticket` parameter still answers 200
      `text/event-stream` and syncs with `me: null` — the spectator path is untouched.
- [ ] A freshly minted ticket still answers 200 `text/event-stream` and its first `sync`
      carries `me.slot` for the player who minted it.
- [ ] `POST /api/rooms/<code>/ticket` with no token answers 400 `bad-token`.
- [ ] `keepSeat` restores `me` from `players` when the held slot is still seated, and
      leaves `me` null when the slot is absent from `players`.
- [ ] **Browser demonstration, before and after, on two games** (Connect 404 and one of
      Tic Tac Overflow / Edge Case). With `redeemTicket` temporarily stubbed to
      `return null` — i.e. redemption fails *every* time, the worst case:
      - before the fix: the seated player's board goes dead, the strip reads
        "You are ⬤ grey", no "Reconnecting…" ever appears and it never recovers;
      - after the fix: the strip shows "Reconnecting…" briefly, then within a few
        seconds the player is back on their own colour with a live board (running on the
        polling fallback), and their opponent never sees them go away.
      Revert the stub afterwards and confirm `git status` is clean of it.
- [ ] `node .agent/scripts/verify-realtime-stream.mjs --base http://127.0.0.1:3155`
      passes every check against `next dev`, **and** against a production build
      (`next build && next start`) — see Risks.
- [ ] `npm run lint` clean.

---

### Verification

```bash
# one-time, in the worktree
docker run -d --name issue155-mysql -e MYSQL_ROOT_PASSWORD=issue155 \
  -e MYSQL_DATABASE=csalinas -p 127.0.0.1:13155:3306 mysql:8.0
npm ci && npx prisma generate
DATABASE_URL='mysql://root:issue155@127.0.0.1:13155/csalinas' npx prisma db push --skip-generate

# dev server
DATABASE_URL='mysql://root:issue155@127.0.0.1:13155/csalinas' ./node_modules/.bin/next dev -p 3155
node .agent/scripts/verify-realtime-stream.mjs --base http://127.0.0.1:3155

# production build — closest to what deploys, and where the bug was also confirmed.
# Stop `next dev` first: never run both against the same worktree, they share .next.
DATABASE_URL='mysql://root:issue155@127.0.0.1:13155/csalinas' npm run build
DATABASE_URL='mysql://root:issue155@127.0.0.1:13155/csalinas' PORT=3156 npm run start
node .agent/scripts/verify-realtime-stream.mjs --base http://127.0.0.1:3156

npm run lint
docker rm -f issue155-mysql
```

Browser check: `http://127.0.0.1:3155/games/connect-404` in two browser profiles (a
normal window and a private one — one identity per browser storage), host one, join the
other with the code, then repeat the stubbed-`redeemTicket` demonstration above.

**Reverting the temporary stub:** commit your real work *first*, then
`git checkout -- src/lib/realtime/tickets.js` — name that one file, never a directory,
and never `git stash` (the stash is shared between agent worktrees).

---

### Risks

- **Already de-risked, keep it verified:** the fix only makes sense if one `tickets.js`
  module instance is shared by the ticket route and the stream route in a production
  build — otherwise every stream would 403 after this change and every online game would
  run permanently on the 2s polling fallback. I confirmed the sharing on a real
  `next build && next start` (a freshly minted ticket redeemed on the first request
  after boot), which is why the production run of the harness is an acceptance
  criterion rather than an open question. **If it ever fails there, stop and report it** —
  do not restore the anonymous fallback to make it pass; that is the bug.
- `keepSeat` re-attaching to a *recycled* slot: in a lobby, if the host removes you and
  somebody else takes the freed slot, an identity-less payload could show you as that
  seat. Only reachable through a transport that has no identity — which change 2
  eliminates — and the consequence is cosmetic, because the server authorizes by token
  and would answer such a move with 403/422. Worth one line of comment, not a mechanism.
- Anything that reads `me === null` as "I am a spectator" now sees it less often. That
  is the fix, but check `edge-case/online/OnlineGame.jsx:240` (`if (spectating || !me)`)
  still behaves: a genuine spectator has no seat in `players`, so `keepSeat` never fires
  for them.
- `applyRoom` is on the hot path for every payload of every game. Keep `keepSeat` an
  O(players) find that allocates nothing when it does nothing (return the same object),
  and keep it out of the revision guard's way.

csalinas-dev:
Closes #155

Both halves of the architect's plan. They fix different things and neither covers the other: the route fix is the only thing that restores **presence**, and the client fix is the only thing that protects the seat from any *future* identity-less payload.

## What changed

1. **`src/lib/realtime/errors.js`** — new `badTicket()` factory (403 `bad-ticket`), beside the other 403s. Its comment records *why* an unredeemable ticket is not downgraded to anonymous.
2. **`src/app/api/rooms/[code]/stream/route.js`** — an **absent** ticket still means spectator, byte for byte. A ticket that is *present* and does not redeem now returns `toResponse(badTicket())` **before any stream is constructed**. The handler's doc comment no longer claims the ticket "only decides who `me` is" — that assumption is what made this silent.
3. **`src/app/api/rooms/[code]/ticket/route.js`** — a tokenless `POST` is now `400 bad-token` up front. Previously it minted a ticket that stood in for nobody, which after change 2 would 403 one request later and read as the room rejecting you. No in-repo caller does this (`useRoom` only mints when it has a token).
4. **`src/lib/realtime/identity.js`** (new, pure, no imports) — `keepSeat(payload, heldSlot)`, the one place that decides when `me: null` is believable, the way `absence.js` is the one place absence is worded. A payload with `me: null` may only un-seat you if your slot is **also gone from its own `players`**. The restored seat is taken from that payload's roster, so `name`/`color`/`connected`/`left` stay fresh.
5. **`src/lib/realtime/useRoom.js`** — a `seatRef`, and every payload routed through `keepSeat` inside `applyRoom`, the single funnel all of them pass through (SSE `sync`, polling tick, `send`'s 409/422 body, join, leave, refresh). Clearing the ref when the merged payload really has no seat is what keeps a lobby leave and a host removal landing.
6. **`src/app/(pages)/games/edge-case/online/OnlineGame.jsx`** — comment only. See "Notes for the reviewer".
7. **`.agent/scripts/verify-realtime-stream.mjs`** (new) — the harness, pure + live, live half run against **two** games.
8. **Docs** — `src/lib/realtime/README.md` (status table, stream row, ticket paragraph, new "Identity in a payload" section) and three `CLAUDE.md` bullets.

## How it was verified

Real MySQL in Docker per the plan's recipe, on **both** `next dev` (:3155) and a **production build** (`next build && next start`, :3156).

### The harness goes RED before it goes green

Proven twice, by reverting each half separately with the other intact (work committed first; reverted with `git checkout -- <one named file>`, never `git stash`):

    # route half reverted to `redeemTicket(ticket, code) || undefined`
    FAIL  a REPLAYED ticket is refused with 403 bad-ticket   expected 403, got 200   (connect-404)
    FAIL  an unknown ticket is refused with 403 bad-ticket   expected 403, got 200   (connect-404)
    FAIL  a REPLAYED ticket is refused with 403 bad-ticket   expected 403, got 200   (tto)
    FAIL  an unknown ticket is refused with 403 bad-ticket   expected 403, got 200   (tto)
    4 checks FAILED.  EXIT=1
    # both anchors still passed, so the failure is the bug and not a broken setup

    # client half neutered (keepSeat returns the payload unchanged)
    FAIL  restores `me` when the held slot is still in `players`
    FAIL  slot 0 is a held seat, not a falsy one       expected 0, got undefined
    FAIL  returns a copy rather than mutating the payload
    EXIT=1

### Green, on both servers

    node .agent/scripts/verify-realtime-stream.mjs --base http://127.0.0.1:3155   # next dev
    node .agent/scripts/verify-realtime-stream.mjs --base http://127.0.0.1:3156   # next build && next start
    -> 18 checks, all ok, EXIT=0 on both

    npm run lint  ->  No ESLint warnings or errors

**The plan's stated risk is retired:** a freshly minted ticket redeems on the first request after `next start`, so the ticket route and the stream route really do share one `tickets.js` module instance in a production build. Online games do **not** fall back to permanent polling.

### Recovery demonstrated, not just the refusal

With `redeemTicket` stubbed to `return null` — redemption fails **every** time, the worst case — on both `connect-404` and `tto`, watching past `PRESENCE_STALE_MS` (45s) and asking the **opponent** what they see, because the presence half is only visible from the other side of the table:

    BEFORE (buggy route + stub) — tto
      stream: 200 text/event-stream
      t=0s    A's own view: me = never established; B sees A connected=true
      t=52s   A's own view: me = null;              B sees A connected=false   (transport: sse (anonymous))

    AFTER (fixed route + stub) — tto
      stream attempt 1: 403 bad-ticket
      stream attempt 2: 403 bad-ticket
      stream attempt 3: 403 bad-ticket
      3 failures -> startPolling(), 2s with X-Player-Token
      t=52s   A's own view: me = seated slot 0;     B sees A connected=true    (transport: polling)

`connect-404` behaved identically. So the worst case is now ~7s of "Reconnecting…" (1s + 2s + 4s backoff) ending in a live board on the polling fallback, with the opponent never seeing the player go away — and the recoverable cases (expiry, replay, a restart, a dev recompile) self-heal on the *first* retry with a freshly minted ticket, well before that.

The stub was reverted with `git checkout -- src/lib/realtime/tickets.js`; `git status` is clean and a grep for stub markers across `src/` and `.agent/` finds nothing.

## Notes for the reviewer

- **I could not drive a real browser** — browser tools were not available in this agent's context — so the plan's two-game *browser* demonstration is the one acceptance criterion I did not perform as written. I substituted the protocol-level before/after above, which measures the exact values that drive those UI strings, and traced them to the components: `"Reconnecting…"` renders on `!connected` (`connect-404/online/OnlineBar.jsx:153`, `tic-tac-overflow/online/OnlineBar.jsx:142`) and the reported "You are ? grey" is `playerFor(players, null)`'s fallback (`connect-404/players.js:88-93`). Worth a human eyeball before merge, but the mechanism is measured rather than reasoned about.
- **Edge Case is comment-only, deliberately, and I want to flag the judgement.** The plan put refactoring `removed`/`heldSlot` out of scope and said do not delete it. Now that the core repairs `me`, that block's roster test (`!players.some(...)`) is arguably redundant — `me` only goes null when the seat is genuinely gone — so it is technically two statements of one rule, which is the #143 hazard the task warned about. I left the code alone and wrote the comment to say the core is authoritative and the local test is a *restatement, not a second rule*, so nobody changes one without the other. The genuinely load-bearing part (`leaving`, which tells *being removed* apart from *standing up*) is untouched and is a question the core cannot answer. Happy to delete the redundant clause if you'd rather have it gone than commented.
- `keepSeat` returns the **same object** when it does nothing (asserted), so it allocates nothing on the hot path that every payload of every game crosses.
- Recycled-slot caveat: in a lobby, an identity-less payload could in principle re-attach you to a slot somebody else has taken. Only reachable through a transport with no identity — which change 2 eliminates — and the server authorizes by token, so such a move gets 403/422. One comment, no mechanism, per the plan.
- Genuine spectators are unaffected: they hold no seat, so `keepSeat` never fires (covered by a check, including the `slot 0` falsy trap).
- The plan's dev-only amplifier is real and I hit it: in `next dev`, the first stream open after the route compiles loses the ticket Map, so the harness compiles the stream route with one bare spectator open before minting (commented at the call site). It does not happen in a production build, and after this change it is a loud 403 with a one-retry recovery like any other lost ticket.
- `npm run lint` prints a Next.js deprecation notice about `next lint` on every run; pre-existing and unrelated.


csalinas-dev:

One blocking finding. Everything else in this PR I verified myself and it holds up — including the two things the PR flagged as unverified. Read the "Verified" section first: the fix is right, and the round trip is for one small change inside the new file.

### Blocking

**1. `src/lib/realtime/identity.js` — `keepSeat` matches on slot number alone, so a recycled seat re-seats you as somebody else, permanently**

`keepSeat` restores `me` from `payload.players.find(p => p.slot === heldSlot)`. Nothing ties that roster entry to *you*. When a lobby seat is vacated and re-taken, the new occupant gets the same slot (`nextSlot` hands out the lowest free seat), so the restored `me` is the newcomer's seat — name, colour and all.

Reproduced against a running server with real payloads (no hand-built objects), driving the real `keepSeat` through `applyRoom`'s exact seat bookkeeping:

```
victim seated: slot 1
  (host removes slot 1; a newcomer joins and takes the freed slot 1)
after the interruption, five consecutive payloads:
  payload 1: server said me=null; client renders me = slot 1 "Newcomer" (orange)
             | victim still on the roster? false
             | edge-case "you were removed" banner fires? false
  payload 2..5: identical
STICKY: never heals — the victim renders as the newcomer until they reload.
```

Concrete outcome: the removed player sits in the lobby wearing the newcomer's name and colour, is never told they were removed (Edge Case's `removed` needs `!me`, and `me` is no longer null), and it never self-heals — every subsequent payload repeats the repair. On `main` today that same sequence correctly un-seats them and shows the banner, so this is a regression the PR introduces.

**The comment that dismisses this is factually wrong, and it sits in the file designated as the single source of truth for the rule:**

```js
// In a lobby a freed slot can be taken by somebody else, so in principle this
// could re-attach to a seat that is no longer yours. It is only reachable
// through a transport with no identity, the consequence is cosmetic, and the
// server authorizes by token — it would answer such a move with 403.
```

Both load-bearing clauses fail:

- *"Only reachable through a transport with no identity"* — no. The victim's transport is fully identified. Their own `X-Player-Token` snapshot legitimately returns `me: null`, because they genuinely hold no seat any more. That is the correct server answer; `keepSeat` is what turns it into the wrong client state. The plan inherited this premise and accepted the risk on the strength of it, which is why I am raising it rather than waving it through.
- *"Needs a sub-second race"* — no, and this is what moves it from cosmetic to blocking. `watch.js` polls the row every `STREAM_POLL_MS` (1s) and publishes only the **latest** revision, so intermediate revisions are routinely collapsed even on a healthy SSE stream. And any stream reconnect — a phone waking, a backgrounded tab, **or the new 403 retry this PR itself adds** — resumes from `getRoom`'s current snapshot, so a removal and a re-join separated by minutes still arrive as one payload. The realistic case is "the host cleared your seat in the lobby while your phone was asleep, somebody else took it, you woke up".

That the consequence is non-destructive (the server still authorizes by token) is true, and is why this is one round trip rather than an architecture question.

**Fix — the discriminator already exists on the wire.** `connectedAt` is in `publicPlayer`'s allow-list, so it is on every seat of every payload including identity-less ones. I checked it does the job:

```
victim seat connectedAt   : 2026-08-16T02:15:52.068Z
newcomer slot1 connectedAt: 2026-08-16T02:15:52.335Z   distinguishable? true

own connectedAt across 4 payloads stable?               true
same seat in an IDENTITY-LESS payload roster:           identical
discriminator survives the legitimate repair?           true
```

So: hold the seat **object** in `useRoom`'s `seatRef` instead of the slot number, and have `keepSeat(payload, heldSeat)` restore only when the roster entry at that slot is the same seat (equal `connectedAt`). It leaves the case this PR exists to fix untouched — verified above, `connectedAt` is stable across payloads and identical in the bare payload — and declines exactly the recycled one. Degrade to today's slot-only behaviour when either `connectedAt` is null, so nothing gets worse if it is ever absent. Please add a case to the pure half of the harness, and correct the comment.

---

### Verified

Real MySQL in Docker, on `next dev` (:4157) **and** on a production build (`next build && next start`, :4158), against **the branch merged with current `origin/main`** — the merge is clean, and `main` has moved on since the branch was cut (Race Condition is now in the registry), so this is the post-merge state rather than the branch in isolation.

**The two-game browser demonstration the PR could not perform — done, and it holds.** Headless Chrome via `.agent/scripts/shot.mjs`, holding a real seat, with a scripted opponent in the other one, sampling the rendered DOM every 2s.

- **True BEFORE** (both halves reverted + `redeemTicket` stubbed to fail every time), connect-404 — the issue's reported symptom, verbatim, and it never recovers:
  `Room 8396  You are grey vs your opponent …`, with `reconnecting: false` at every sample from t=0 to t=18s.
- **AFTER** (both halves + the same stub), connect-404: `reconnecting: true` at t=0, then from t=2s `You are Blue` with a live board, running on the polling fallback. Recovery, not just refusal.
- **AFTER**, tic-tac-overflow, run past `PRESENCE_STALE_MS` and watched from the **opponent's** side — the half `keepSeat` cannot fix:
  `t=15…75s: ["0:connected","1:connected"]` throughout. The screenshot shows `Room 6JYP  You are ⭘ vs Opponent` with a live board.
- **Route half reverted, `keepSeat` intact, same stub** — the run that proves both halves are independently load-bearing, and worth recording: the player looks *completely fine to themselves*, but the opponent sees `["0:connected","1:AWAY"]` from t=10s to t=85s, permanently. The client half alone hides the symptom while leaving the player counted out by everyone else at the table.

**The harness goes RED for each half separately** (each reverted with the other intact, anchor-checked before mutating, restored with a named-path checkout):

- route half only → 4 live FAILs across connect-404 and tto, both anchors still green.
- `keepSeat` only → 3 pure FAILs, including the `slot 0` falsy trap.

**The PR's biggest stated risk is genuinely retired.** `verify-realtime-stream.mjs` is 18/18 on `next dev` *and* on the production build, with the anchor "a fresh ticket opens an SSE stream as slot 0" passing on the first request after `next start`. The ticket and stream routes really do share one `tickets.js` instance in production; games do not fall back to permanent polling.

**Regression on the other games, which is the real danger here:**

- The absent-ticket spectator path is byte-for-byte the old behaviour — `token` is simply left `undefined` and `getRoom(code, undefined)` is reached exactly as before. Green in the harness on both servers.
- `rc-room.mjs check --presence` against the production build: **35/35**, 40 response bodies scanned for tokens — spectator-on-a-full-room, third arrival, mid-game leave and rejoin, stale revisions, the rematch, and presence-as-absence.
- All four games joined and rendered in a browser: connect-404, tic-tac-overflow, race-condition (`You are G Gold vs Cyan`), and the edge-case lobby (seat, colour lock, 2/4 roster).
- A real move exchange on the production build: the scripted opponent's drop reaches the browser player — `1 / 42 pieces … Opponent dropped into column 1 … Player 2 to move`.
- No other caller of `POST /ticket` exists anywhere in `src/` or `.agent/`, and `useRoom` only mints when it holds a token, so the new 400 closes a hole rather than changing behaviour.
- `npm run lint` clean.

**`keepSeat` attacked with real server payloads** — the cases you would expect to break it do not:

- lobby host-removal really un-seats the victim (`held 1 -> me null`)
- lobby self-leave really un-seats
- mid-game self-leave: the server still resolves `me` by token (`me {slot:1,left:true}`), so `keepSeat` never fires — the seat that must survive does
- an expired room is a 410 carrying no payload, so nothing reaches `keepSeat` at all

Only the recycled-slot case broke through.

### Non-blocking

- **Edge Case's `removed` clause — my call: keep it, and tighten one line of the comment.** You asked for a decision. The comment frames the whole test as "a restatement of `keepSeat`'s rule". Half of it is; the other half is not. `heldSlot.current !== null` is what separates *I had a seat and lost it* from *I never had one*, and the core deliberately does not answer that — `keepSeat` repairs `me`, it does not tell you that you were removed. Delete that half and a genuine spectator gets "you were removed". So the clause stays, and the comment should name which half is the restatement (`!players.some(...)`) rather than implying the block as a whole is. This is also the opposite of the #143 shape: the local test can only ever make the banner fire *less*, and the blocking finding above is a live demonstration that the core invariant is not currently airtight enough to lean on unguarded.
- `POST /ticket` guards with `!body.token` but never runs `assertToken`, so `{"token": 1}` or `{"token": "x"}` still mints a ticket that redeems to a token holding no seat — i.e. a 200 anonymous stream, hand-crafted only. No client path reaches it, but `assertToken` would be symmetric with every other route if you are in the file anyway.
- The documentation quality throughout is genuinely high — `errors.js`'s `badTicket` comment and the README's "Identity in a payload" section both explain *why* rather than *what*, and I navigated by them. That is precisely why the one wrong sentence in `identity.js` is worth correcting rather than shrugging at: everything around it is trustworthy enough that the next reader will believe it.

csalinas-dev:
Addressed, in `86f64ac`.

## Blocking — `keepSeat` matched on slot alone

**You were right on both clauses, and the comment was the reason this got past a plan and a review.** It is rewritten to say what is actually true: the victim's transport is fully identified — their own `X-Player-Token` snapshot legitimately says `me: null` because they genuinely hold no seat — and no sub-second race is needed, since the 1s watcher publishes only the latest revision and any reconnect (including this PR's own 403 retry) resumes from the current snapshot.

**Fix, using `connectedAt` exactly as you suggested — no new field.**

- `useRoom`'s `seatRef` now holds the **seat object**, not its slot (`seatRef.current = merged.me ?? null`).
- `keepSeat(payload, heldSeat)` finds the roster entry at `heldSeat.slot` and restores only when it is the *same seat*: if both sides carry a `connectedAt` and they differ, it refuses and lets the removal land. With `connectedAt` missing on either side it degrades to today's slot-only match, so nothing gets worse if it is ever absent.
- `connectedAt` is stamped once when a seat is created and never rewritten (`lastSeenAt` is the field that moves — checked in `rooms.js`'s rejoin path, which touches `lastSeenAt` and `left` and leaves `connectedAt` alone), which is what makes it stable across the payloads the legitimate repair needs.

### Proved RED first, with real server payloads

Your reproduction, scripted against a running server (Edge Case, because it has a real lobby a host can remove from): host creates, victim joins slot 1, host removes slot 1, newcomer joins and `nextSlot` hands them the same slot 1 — then the victim's own authenticated snapshot, five times, through `applyRoom`'s exact bookkeeping.

    pre-fix (keepSeat holding a slot number)
      anchor: B is really seated       -> slot 1, connectedAt 2026-08-16T02:33:37.996Z
      host removes slot 1              -> 200
      newcomer takes the freed seat    -> slot 1, connectedAt 2026-08-16T02:33:38.202Z
      connectedAt distinguishable?     -> true
        server me=null -> client 1:Newcomer     (x5, identical)
      RED  — the removed player renders as the newcomer, and it never heals.

    post-fix (keepSeat holding the seat object)
        server me=null -> client null           (x5)
      GREEN — the removed player stays un-seated; the removal lands.

Same probe, same result, on `next dev` **and** on the production build. Note the two seats were 206ms apart even with back-to-back scripted requests.

The probe detects which version of `keepSeat` it is driving and adapts its bookkeeping, so the RED is the *rule* failing, not a signature mismatch. Worth saying because the harness's own new case cannot show that: run the new pure half against the old `identity.js` and 3 of 8 fail on the object-vs-number mismatch while the recycled case "passes" for the wrong reason. That is why the RED above is the probe rather than the harness.

### Harness

Two pure cases added (now 8 pure + 7 live per game = **22**):

- `declines a RECYCLED slot: the newcomer in your old seat is not you` — three consecutive payloads, so stickiness is asserted, not just the first render.
- `falls back to the slot when connectedAt is missing` — both directions (absent on the roster entry, absent on the held seat).

All the existing pure cases now pass seat objects; the live half is unchanged apart from one addition below.

## Non-blocking

- **Edge Case's `removed` clause — kept, comment corrected.** It now says `heldSlot.current !== null` is what separates *I had a seat and lost it* from *I never had one* (the question the core deliberately does not answer, and what stops a genuine spectator being told they were removed), and that only the `players.some(...)` half restates `keepSeat` — **and more coarsely**, since the core also declines a slot a newcomer has taken. Where they differ the local test simply does not fire, which is the safe direction: it can make the sentence appear less often, never wrongly. (That is unchanged from `main`, which does not fire there either.)
- **`assertToken` on `POST /ticket`.** Done — `assertToken` is now exported from `rooms.js` and the route runs it instead of `!body.token`, same `bad-token` 400. New harness case: `POST /ticket` with `{"token":"x"}` → 400 `bad-token`.

## Verified

Real MySQL in Docker, throwaway container, `next dev` (:5155) and a production build (`next build && npm run start`, :5156).

    node .agent/scripts/verify-realtime-stream.mjs --base http://127.0.0.1:5155   22/22
    node .agent/scripts/verify-realtime-stream.mjs --base http://127.0.0.1:5156   22/22
    npm run lint                                                                  clean

The live half is re-run rather than taken on trust because this pass touched a server file (`ticket/route.js`, plus the `assertToken` export): fresh tickets still mint and redeem on both servers, so the shared-`tickets.js` result you confirmed still holds.

Headless Chrome against the production build, both games, holding a real seat in a room created over the API:

- connect-404 — `Room PVG7  You are Blue vs Host`, live board, `grey: false`, `reconnecting: false`
- tic-tac-overflow — `Room BZ8N  You are ⭘ vs Host`, live board, same

That is a smoke test of the client path under the new `seatRef`, not a re-run of your before/after demonstration — I have not re-done anything you confirmed (both halves independently load-bearing, the byte-for-byte spectator path, the four-game join, the presence run).

**One limit, stated rather than hidden:** the discriminator is a millisecond ISO timestamp, so a seat vacated and re-taken *within the same millisecond* would collide and fall back to the slot match. It needs a removal and a fresh join to land in the same millisecond over HTTP; I measured 206ms and 267ms (yours) on back-to-back scripted requests. If you want that closed rather than noted, it wants a per-seat id on the row, which is a schema change and outside this issue.

Temporary reverts were done with a named-path `git checkout` after committing, never `git stash`; the throwaway MySQL container and both servers are torn down and the tree is clean.

csalinas-dev:
## 🤖 Review: APPROVED

Re-review of `86f64ac`, scoped to the revision. I did not re-derive what the previous pass confirmed (both halves independently load-bearing, the byte-for-byte spectator path, four games joining, the browser before/after). I graded five things and ran all of them.

### 1. Does `connectedAt` actually identify the seat? — yes, and the degradation is unreachable

Verified structurally and empirically, against real server payloads on a production build:

- `publicPlayer`'s allow-list emits `connectedAt` unconditionally (`players.js:20`), so the key is on **every** seat of **every** payload, identity-less ones included. My probe confirmed it on a bare tokenless snapshot: roster `["…28.360Z","…28.536Z"]`, and the same seat carries an **identical** `connectedAt` in the identity-less payload as in the owner's authenticated one — which is what the legitimate repair depends on.
- Both stamping sites (`rooms.js:156` create, `rooms.js:265` join) build the player object field-by-field from validated inputs. Nothing spreads a request body, so **no client can inject a seat with `connectedAt: null`**.
- `git log -S connectedAt` puts the field in `49e0547`, the first realtime commit. There are no legacy rows that predate it, and rooms 410 after 60 minutes anyway.
- The held side only ever comes from `merged.me`, which is also `publicPlayer` output.

I then **forced** the degradation rather than reasoning about it — stripped `connectedAt` off a real payload's roster, and separately off the held seat:

```
degradation: roster without connectedAt    -> me = "Newcomer"
degradation: held seat without connectedAt -> me = "Newcomer"
```

So the degradation genuinely *is* the old bug — it is safe because it is unreachable, not because it is benign, and the code is explicit and commented about that. The direction chosen (degrade toward restoring rather than toward refusing) is also the right one on merit: refusing would re-open #155, a live production defect, while restoring re-opens a cosmetic, non-authorizing one. Not a finding.

### 2. The millisecond limit — acceptable to ship, no schema change

Agreed with the implementor, and I put numbers on it. A collision needs the **same slot** vacated and refilled inside one millisecond — two HTTP round trips and two read-modify-write commits between the two stamps. Measured on a warm production build with back-to-back scripted requests and no network latency:

```
tightest observed vacate->refill connectedAt gap:  42ms   (next start)
                                                  165ms   (next dev)
```

~42x the window at machine speed; human-driven it is seconds. Two seats sharing a millisecond on *different* slots is harmless — the slot lookup runs first. And a collision degrades to exactly `main`'s behaviour today: cosmetic, the server still resolves seats by token so a move gets 403/422, and it clears on reload. A per-seat id would be a schema change riding `deploy.yml`'s `db push` for a defect with no reachable trigger. Correctly scoped out; the README/PR note is the right level of record.

### 3. The harness caveat — it discriminates, and more strongly than you claimed

You are right that the *old* file cannot be the RED baseline (signature change), but you understated your harness. I ran the signature-fair reversion: with the object signature intact and **only** the `connectedAt` guard removed from `identity.js` (anchor-asserted verbatim before mutating), the committed harness fails **exactly one** case:

```
  ok    restores `me` when the held seat is still in `players`
  ok    leaves `me` null when the held seat is gone from `players`
  FAIL  declines a RECYCLED slot: the newcomer in your old seat is not you
          a removed player was re-seated as the newcomer
          expected null
          got      {"slot":1,"name":"Newcomer","connectedAt":"2026-08-16T02:15:52.335Z"}
  ok    falls back to the slot when `connectedAt` is missing
  ok    (4 more pure)   ok (14 live)
EXIT=1
```

Everything else stays green, so the failure is the **rule**, not the setup. That is a real RED, and the "test that cannot fail against its own bug" pattern does not apply here.

Independently reproduced end to end with real payloads through `applyRoom`'s exact bookkeeping (host creates, victim takes slot 1, host removes slot 1, newcomer recycles slot 1):

```
victim   connectedAt 2026-08-16T02:47:34.276Z
newcomer connectedAt 2026-08-16T02:47:34.335Z
ok  the removed player stays un-seated across 5 payloads (their own token)
ok  an identity-less payload does not re-seat the removed player as the newcomer — me -> null
ok  REGRESSION GUARD: an identity-less payload does NOT un-seat a seated player — me -> 1
```

That last line matters as much as the fix: the case this PR exists for still works.

### 4. `assertToken` on `POST /ticket` — correct, and it belongs here

Correct, and correctly *weak*: `assertToken` is format-only (`TOKEN_PATTERN`). Requiring the token to be **seated** would have been the tempting overreach and would break a spectator who still has a token in `localStorage` from an earlier game — they legitimately mint a ticket and legitimately get `me: null`. This is a strict tightening of a route that previously minted a ticket for `{"token": 1}`; `useRoom` is the only in-repo caller and always holds a real token. The export is two lines on a helper that route already imports (`getRoom`), so no new coupling and no bundle reach.

It belongs in this PR: it was the previous pass's own non-blocking item, it is in a file this PR already rewrites, and splitting it would mean a second round trip through the same server route for four lines. Harness case added and passing on both servers.

### 5. Harness green on both, post-merge

Run against the branch **merged with current `origin/main`** (`262e181`; the merge is clean and main has moved since the branch was cut), real MySQL in a throwaway container:

```
node .agent/scripts/verify-realtime-stream.mjs --base http://127.0.0.1:3255   # next dev             22/22
node .agent/scripts/verify-realtime-stream.mjs --base http://127.0.0.1:3256   # next build && start  22/22
npm run lint                                                                  clean
```

Both anchors pass on the first request after `next start`, so the shared-`tickets.js` result still holds after this pass touched `ticket/route.js`.

**Client half smoke, real headless Chrome against the production build** — because this revision changed `useRoom`, not just the pure module. A real browser joining `?room=9JNQ` on connect-404 held its seat under the new `seatRef` across 8 samples:

```
t=1.5s … Room 9JNQ  You are Blue vs Host … Host to move
t=12s  … Room 9JNQ  You are Blue vs Host … Host to move
```

No "grey", no "?", stable for 12s. (`seatRef` is read in exactly one place — `keepSeat(next, seatRef.current)` — and Edge Case's `heldSlot` is a separate ref that is still a slot number and untouched, so there is no stale-type reader anywhere.)

### Acceptance criteria

- [x] A recycled lobby seat no longer re-seats the removed player — RED proved with a signature-fair reversion, GREEN with real payloads on both servers
- [x] `connectedAt` is the discriminator, present on every real payload, degradation path unreachable by construction
- [x] `POST /ticket` runs `assertToken`; `{}` and `{"token":"x"}` are both 400 `bad-token`
- [x] Edge Case's comment names which half restates `keepSeat`
- [x] Harness 22/22 on `next dev` and on a production build; lint clean

### Non-blocking

- **The PR body is stale.** It still describes `keepSeat(payload, heldSlot)` and still carries the "Recycled-slot caveat" bullet whose premise this revision disproved. `identity.js`, `README.md`, `CLAUDE.md` and the commit message are all correct, but the PR body is what a human opens first. Worth a one-minute edit before merge.
- `ticket/route.js`'s new comment says a malformed token "would mint a ticket that redeems to a seatless stream" — true, but a well-formed random UUID does exactly the same and is still (correctly) accepted. The sentence slightly oversells what `assertToken` buys; the guard is right, the reason is narrower than stated.
- Edge Case's `removed` banner still does not fire in the recycled case (`players.some(slot === 1)` is true, because the newcomer is sitting there). Your comment says so, and I confirmed `main` behaves identically — a pre-existing gap, not a regression, and outside this issue.

Throwaway MySQL container and both servers torn down; the probe mutation was anchor-asserted before it was made and reverted with a named-path checkout; worktree clean.

PR #158, opened 2026-08-16
fix(tooling): make shot.mjs fail when the page script says it failed

### LINKED ISSUE #150: fix(tooling): shot.mjs exits 0 even when a page script reports ok: false
Found by the reviewer of #144 while verifying PR #147.

`.agent/scripts/shot.mjs` exits 0 regardless of what a page script passed via `--eval` reports. The reviewer ran a deliberately broken build that scored **24/24 board disagreements** — every marble on screen wrong — and `shot.mjs` still exited 0.

### Why it matters more than it looks

The verification scripts this repo has grown (`rc-room.mjs park --eval`, and anything future agents build on the same pattern) hand `shot.mjs` a page script that returns a pass/fail result. Because the exit code ignores that result, **every one of those is a report that a human has to read, not a gate that fails**. An agent that runs one, sees exit 0, and reports success is doing something reasonable and getting the wrong answer — and it is exactly the shape of the false-greens this pipeline has caught by hand four times in two days (#134 was another `shot.mjs` one).

There is no CI here, so these scripts are the only automated gate that exists. A gate that cannot fail is worse than no gate, because it is trusted.

### Done when
- A page script reporting `ok: false` (or throwing) makes `shot.mjs` exit non-zero.
- The failing result is printed, not just the exit code — a bare non-zero teaches nobody anything.
- Existing callers are checked: anything that currently relies on exit 0 while reporting failures internally must be updated in the same change, or it will start failing correctly and look like a regression.
- Note in the repo's CLAUDE.md that the exit code is now meaningful, since agents have been told to read the output instead.

---

## 🤖 Implementation Plan

_Written by the architect agent on 2026-08-16 01:46 UTC. The implementor builds exactly this._

### Goal

`.agent/scripts/shot.mjs` becomes a gate instead of a report: when the page script
handed to it via `[evalFile.js]` reports `ok: false`, throws, or returns no value at
all, the process exits **3** and prints what came back. A plain measurement script
(a result object with no `ok` field) keeps exiting 0, but says out loud on stderr
that it asserted nothing, so "no `ok`" cannot quietly become the new false green.

### Approach

One file does the work (`shot.mjs`); two callers and `CLAUDE.md` are corrected so
nothing still tells a reader the exit code is meaningless.

Three decisions, made deliberately because the issue asks for all three:

| what the page script did | verdict | why |
|---|---|---|
| returned an object with `ok` falsy | **fail, exit 3** | The convention the Race Condition scripts settled on. |
| threw (or its promise rejected) | **fail, exit 3** | An assertion that never finished is not an assertion that passed. Today this exits 1 *by accident* — the top-level `await` rejects unhandled — which also means no payload is printed and no screenshot is taken. Both get fixed. |
| returned `undefined` or `null` | **fail, exit 3** | The tool's whole contract is "prints `result: <the eval file's value>`". A file that yields nothing produced neither a measurement nor an assertion, and the likeliest cause is an async IIFE that forgot to `return`. Escape hatch for a genuinely side-effect-only script: end the file with `true`. |
| returned a value with **no** `ok` field | **pass, exit 0, loud stderr NOTE** | This is the plain-measurement case (`({width: 320, gap: 12})`) that `shot.mjs` was built for and that `CLAUDE.md:106` documents. Failing it would break every layout measurement. Silently passing it is the hole one level up, so it is passed *noisily*: one stderr line saying this run asserted nothing. |

Rejected: requiring an opt-in `--assert` flag to turn on gating. It puts the gate
behind a flag every future caller can forget, which is the same failure with an
extra step. Rejected: treating a missing `ok` as failure — it breaks the
measurement use the tool exists for.

A new exit code **3** rather than reusing 1, because "Chrome/CDP broke" and "the
page is wrong" are different findings and the caller can now tell them apart. Both
are non-zero, which is what the issue asks for.

**The screenshot must still be written on a failing run.** A failed replay is
exactly when you want the PNG. So the verdict is *recorded* where the result is
read, the existing `console.log(payload)` and screenshot block run unchanged, and
the non-zero exit happens at the very end.

- `.agent/scripts/shot.mjs:151-153` — `const payload = { viewport }; if (evalFile) payload.result = await evaluate(...); console.log(...)`. This is the site of the change. `.agent/scripts/shot.mjs:175` — the unconditional `process.exit(0)`.
- `.agent/scripts/shot.mjs:114-128` — `evaluate()` already turns `exceptionDetails` into a thrown `Error` whose message starts `evaluate threw: `. It is called twice: once for the viewport read-back (line 143) and once for the eval file (line 152). **Only the eval-file call changes behaviour** — a throw from the viewport read is the tool failing and must stay exit 1.
- `.agent/scripts/shot.mjs:52-59` — the `process.on("exit")` handler closes the WebSocket and kills Chrome synchronously, so `process.exit(3)` cleans up exactly like `process.exit(0)` does. No teardown work is needed.
- Confirmed on this machine (Node v24.11.0): an unhandled top-level-await rejection exits **1** and dumps a stack. That is today's behaviour for a throwing page script, and it prints no payload and takes no screenshot.
- Prior art for this exact kind of fix, including commit-message voice: `git show 8811478` ("fix(tooling): make shot.mjs fail loudly instead of measuring the wrong viewport (#135)"). Follow it — root cause, not symptom, and the comment above each guard says which failure it catches.
- This repo has **no test runner and no CI**. `.github/workflows/deploy.yml` is the only workflow and it never touches `.agent/scripts`. Nothing in `package.json` invokes `shot.mjs`. Verification is by running the script, which is why the verification section below is a matrix rather than a suite.

#### Caller audit — every use of `shot.mjs` on `main`

1. **`.agent/scripts/rc-ending.mjs`** — generates a page script (the `pageScript` template starting at line 243) that replays a whole game by clicking. **Every return path already carries `ok`**: `fail = (error) => ({ ok: false, error })` at line 249, used at lines 302 and 311, and the final return at lines 347-353 computes `ok: played === TURNS.length && finished`. No shape change is needed. **This is the caller that will start failing correctly**: a replay that today prints `ok: false` and exits 0 will, after this lands, exit 3. That is the fix working, not a regression — say so in the PR body. Its comments at lines 31-33, 237-242 and 245 assert the opposite ("shot.mjs exits 0 even when the page script it ran reported a failure (#150)") and are now false; they must be rewritten.
2. **`.agent/scripts/rc-room.mjs` `park --eval`** — the `pageScript` template starting at line 506. Its final return (lines 565-571) carries `ok: disagreed === 0 && moved === 0`, but **two early returns do not carry `ok` at all**: line 542 `return { error: "board never appeared" };` and line 547 `return { error: "never offered a turn" };`. Under the new rule those land in the "no `ok` field" bucket and would exit 0 with a NOTE — a real failure passing. **Both must become `{ ok: false, error: ... }` in this same change.** This is the concrete instance of "silently treating no `ok` as success would rebuild the same hole one level up".
3. **`CLAUDE.md:88`** (the `rc-ending.mjs` bullet) — contains "**Read `result.ok`, not the exit code** — `shot.mjs` exits 0 even when the page script it ran reported a failure (#150)". Now wrong.
4. **`CLAUDE.md:106`** (the `shot.mjs` bullet) — documents the exit-code table: "A bad argument exits **2** ... and a CDP error or failed navigation exits **1**". Incomplete now.
5. **`CLAUDE.md:92`** (the `rc-room.mjs` bullet) — describes `park --eval` writing a page script that asserts board/room agreement. Says nothing about exit codes; a short clause that it is now a gate is welcome but optional.
6. `.agent/scripts/verify-race-condition.mjs` never touches a browser or `shot.mjs`. Leave it alone.
7. No plain-screenshot (`out.png`, no eval file) caller is committed anywhere — that use is ad-hoc from the command line and from agent sessions. It must keep behaving exactly as it does today, and the matrix below has a case pinning that.

### Changes

1. **`.agent/scripts/shot.mjs`**
   - Give `evaluate()` a label parameter so the two call sites are distinguishable in the message, e.g. `const evaluate = async (expression, what = "evaluate") => ...` throwing `` `${what} threw: ${description}` ``. Call the eval-file one as `evaluate(readFileSync(evalFile, "utf8"), "the page script")`; leave the viewport read on the default.
   - Replace lines 151-153 with: build `payload = { viewport }`, then when `evalFile` is present, evaluate it inside `try/catch` and derive a verdict into two locals — `failure` (a string reason, or `null`) and `asserted nothing` (a boolean for the NOTE):
     - catch → `payload.error = e.message` ("the page script threw: …"), `failure = payload.error`.
     - result `undefined` or `null` → `payload.error = "the page script returned no value — nothing was measured and nothing was asserted"`, `failure = payload.error`. (Set `payload.error` explicitly, because `JSON.stringify` drops an `undefined` `result` key entirely and the printed payload would otherwise say nothing at all.)
     - result is a non-null object **and** `"ok" in result` → `failure` is set iff `!result.ok`, and the reason names the reported value and appends `result.error` when the page script supplied one.
     - anything else → no failure, flag the NOTE.
   - Leave `console.log(JSON.stringify(payload, null, 2))` and the whole `if (out && out !== "-")` screenshot block exactly where they are and exactly as they are.
   - Replace the trailing `process.exit(0)` with: when `failure`, print to **stderr** a banner line `shot.mjs: FAILED — <reason>` followed by `JSON.stringify(payload.result ?? payload, null, 2)` — the full result object, re-printed on stderr so a caller reading only stderr or only the tail sees *what* disagreed, which is the thing that let the reviewer of #144 act — then `process.exit(3)`. Otherwise, when the NOTE flag is set, print one stderr line along the lines of `shot.mjs: NOTE the page script returned no "ok" field, so this run asserted nothing — exit 0 means Chrome ran and printed a value, not that anything passed.` and `process.exit(0)`.
   - Update the header comment block (lines 1-5): the exit table becomes `0 ok, 2 bad usage, 1 runtime/CDP failure, 3 the page script reported failure (ok falsy, threw, or returned nothing)`, and one sentence saying a result with no `ok` field is a measurement, not an assertion, and exits 0.
   - Keep the existing viewport-mismatch WARNING at exit 0. It is Chrome behaving correctly; it is not a page-script verdict.
2. **`.agent/scripts/rc-room.mjs`** — lines 542 and 547: add `ok: false` to both early returns. Add one line to the page-script comment block (around lines 490-505) saying every return carries `ok` because `shot.mjs` now exits 3 on a falsy one. Optionally extend the printed hint at lines 616-619 with "shot.mjs exits 3 if the board and the room disagree".
3. **`.agent/scripts/rc-ending.mjs`** — no behaviour change; correct the three comment blocks that now state the opposite of the truth: lines 31-33 (the `READ result.ok FROM THAT SECOND COMMAND. shot.mjs exits 0 even when…` paragraph), lines 237-242, and the generated banner at line 245 (`// Read \`result.ok\` — shot.mjs exits 0 whatever this returns (#150).`). New wording: the replay is a gate — `ok: false` makes `shot.mjs` exit 3 and print the result. **Keep the record of the silent-green incident** (the mode picker in front of the board, `{"error":"board not found"}`, zero turns, exit 0) — it is why the anchor assertions exist; only the claim about the exit code changes.
4. **`CLAUDE.md`** — update line 106's exit table to include 3 and the no-`ok` NOTE; update line 88 to drop "shot.mjs exits 0 even when the page script reported a failure" and say the replay now fails the command. Both bullets should note the "end the file with `true` if you really want a side-effect-only page script" escape hatch is available (one clause on line 106 is enough). Do not restructure the file or reword neighbouring bullets.

### Out of scope

- Any new flag, option, or mode on `shot.mjs` (`--assert`, `--json`, a strict switch). The gate is unconditional.
- Adding a top-level `ok` to the printed payload, or otherwise changing the stdout shape beyond adding `error` on the two failure paths. Existing readers parse `{viewport, result}`.
- Touching the viewport read-back, the WARNING, the 3500ms settle, the screenshot sizing, or the Chrome flags.
- `verify-race-condition.mjs`, the `check` subcommand of `rc-room.mjs`, and anything under `src/`.
- Weakening any page script's `ok` computation to make a run go green. If the rc-ending replay fails after this change, report the failure — do not tune the assertion.
- Adding a test runner or a CI workflow.

### Acceptance criteria

- [ ] `node .agent/scripts/shot.mjs about:blank 900 - <file returning ({ok:false, error:"deliberate"})>` exits **3**, prints the result object on stdout, and prints a `FAILED` banner plus the result object on stderr.
- [ ] The same command with `({ ok: true })` exits **0**.
- [ ] A page script that throws exits **3**, and the printed payload carries an `error` naming the throw — it no longer dies with an unhandled-rejection stack.
- [ ] A page script returning nothing (`void 0;`) exits **3** with an error saying no value came back.
- [ ] A page script returning a value with no `ok` field exits **0** and prints exactly one stderr NOTE saying nothing was asserted.
- [ ] A run with **no** eval file (plain screenshot) exits **0** and writes the PNG, unchanged.
- [ ] A failing run still writes the PNG when an output path was given — the screenshot is taken before the non-zero exit.
- [ ] Bad usage still exits **2**; a dead server / failed navigation still exits **1**.
- [ ] `.agent/scripts/rc-room.mjs` grep for `return {` inside the page-script template shows `ok` on **every** return path.
- [ ] No file in the repo still claims `shot.mjs` exits 0 when a page script fails: `grep -rn "exits 0" .agent CLAUDE.md` surfaces only statements that are true after this change.
- [ ] `CLAUDE.md` documents exit 3 and the no-`ok` NOTE.

### Verification

Write the probe files inside your own worktree with issue-scoped names (e.g.
`shot150-*.js`) and delete them before committing — they must not land in the PR.

    # 1. ok:false -> 3
    printf '({ ok: false, error: "deliberate" })\n' > shot150-false.js
    node .agent/scripts/shot.mjs about:blank 900 - shot150-false.js; echo "exit=$?"   # 3

    # 2. ok:true -> 0
    printf '({ ok: true })\n' > shot150-true.js
    node .agent/scripts/shot.mjs about:blank 900 - shot150-true.js; echo "exit=$?"    # 0

    # 3. throws -> 3
    printf '(() => { throw new Error("boom"); })()\n' > shot150-throw.js
    node .agent/scripts/shot.mjs about:blank 900 - shot150-throw.js; echo "exit=$?"   # 3

    # 4. returns nothing -> 3
    printf 'void 0;\n' > shot150-void.js
    node .agent/scripts/shot.mjs about:blank 900 - shot150-void.js; echo "exit=$?"    # 3

    # 5. measurement, no ok field -> 0 + NOTE on stderr
    printf '({ width: innerWidth })\n' > shot150-measure.js
    node .agent/scripts/shot.mjs about:blank 900 - shot150-measure.js; echo "exit=$?" # 0

    # 6. plain screenshot, no eval file -> 0, PNG written
    node .agent/scripts/shot.mjs about:blank 900 shot150-plain.png; echo "exit=$?"    # 0

    # 7. failing run still writes the PNG
    rm -f shot150-fail.png
    node .agent/scripts/shot.mjs about:blank 900 shot150-fail.png shot150-false.js; echo "exit=$?"  # 3
    ls -l shot150-fail.png                                                            # exists

    # 8. usage and CDP failures unchanged
    node .agent/scripts/shot.mjs; echo "exit=$?"                                      # 2
    node .agent/scripts/shot.mjs http://localhost:59999/ 900 -; echo "exit=$?"        # 1

`about:blank` is deliberate: none of the above needs a server, and every case is
about what the page script returned rather than about the page. Paste the eight
exit codes into the PR body.

Then one real end-to-end run, which is the point of the whole exercise — a caller
that passes must still pass. Build and serve on a port derived from this issue so
it cannot collide with another agent's server (never run `next dev` against a
worktree a `next start` is serving):

    npx next build
    npx next start -p 3150 &
    node .agent/scripts/rc-ending.mjs orbited-in --eval shot150-ending.js
    node .agent/scripts/shot.mjs http://localhost:3150/games/race-condition 900 shot150-ending.png shot150-ending.js; echo "exit=$?"   # 0

Kill that server by the PID listening on **3150** (`netstat -ano | grep ':3150.*LISTENING'`
then `taskkill //F //PID <pid>`), never by image name.

If `next build` fails for a reason unrelated to this change, say so explicitly in
the PR body rather than dropping the end-to-end run silently. The `rc-room.mjs park`
half needs a live `DATABASE_URL` and is optional; if you can run it, do, and report
the exit code.

### Risks

- **A currently-red caller turning red.** `rc-ending.mjs`'s replay computes `ok`
  honestly today and its failures are invisible. The first person to run it after
  this merges may see a non-zero exit and read it as this PR breaking something.
  Pre-empt it in the PR body: name `rc-ending.mjs` and `rc-room.mjs park --eval` as
  the two scripts whose *existing* failures now surface, and state that step 6 of
  the verification above shows the passing path still exits 0.
- **Flaky page scripts become hard failures.** The replay clicks a real board with
  fixed sleeps. If it is timing-sensitive it will now fail the command rather than
  print a quiet `ok: false`. That is the intended trade — the fix is to make the
  page script wait properly (it already waits for hydration), never to soften `ok`.
- **An ad-hoc eval file that returns nothing.** Any hand-written page script an
  agent wrote in a past session that ends on a statement rather than an expression
  now exits 3. Cheap to fix (end the file with `true`) and documented in
  `CLAUDE.md`; nothing committed in this repo does it.
- **Exit code 3 mistaken for a crash.** Mitigated by the stderr banner naming the
  page script's own reason, and by the exit table in the header comment and
  `CLAUDE.md:106`.

csalinas-dev:
Closes #150

## What changed

- **`.agent/scripts/shot.mjs` — the page script's verdict is now the exit code.** The result of `[evalFile.js]` is read into two locals (`failure`, `assertedNothing`) at the point it is evaluated, and acted on at the very end:

  | what the page script did | exit |
  |---|---|
  | returned a value with falsy `ok` | **3** |
  | threw / its promise rejected | **3** |
  | returned `undefined` or `null` | **3** |
  | returned a value with **no** `ok` field | **0**, plus one loud stderr `NOTE` |
  | no eval file at all | **0**, unchanged |

  On failure it prints `shot.mjs: FAILED — <reason>` on stderr followed by the whole result object, again on stderr. A bare non-zero teaches nobody anything — *what* disagreed is the part that let the reviewer of #144 act, so it is repeated where a caller reading only stderr, or only the tail, will see it.

- **The verdict is recorded, not acted on immediately.** `console.log(payload)` and the entire screenshot block are untouched and still run, so a failing run still prints its payload and still writes its PNG. A failed replay is exactly the run you want the screenshot of.

- **The throw path is genuinely new behaviour, not just a new number.** Previously a throwing page script died as an unhandled top-level-await rejection: exit 1 by accident, a Node stack instead of a payload, and no screenshot. It now lands in the payload as `error` and exits 3 with the PNG written.

- **`evaluate()` takes a label.** The eval-file call is `evaluate(..., "the page script")`; the viewport read-back keeps the default. A throw from the viewport read is the *tool* failing and still exits 1 — the two are different findings and a caller can tell them apart.

- **`.agent/scripts/rc-room.mjs` — both `ok`-less early returns fixed.** `{ error: "board never appeared" }` and `{ error: "never offered a turn" }` are real failures that would have landed in the new "no `ok`" bucket and passed. Both now carry `ok: false`. Fixing `shot.mjs` without this ships a gate with a hole already in it. The page-script comment block says why every return must carry `ok`, and the printed hint now mentions exit 3.

- **`.agent/scripts/rc-ending.mjs` — three stale comment blocks corrected** (lines 31-33, 237-242, and the generated banner). No behaviour change: every return path there already carried `ok`. **The record of the silent-green incident is kept** — the mode picker in front of the board, `{"error":"board not found"}`, zero turns, exit 0 — because it is why the anchor assertions exist. Only the claim about the exit code changed.

- **`CLAUDE.md` lines 88 and 106** — the exit table gains 3 and the no-`ok` NOTE, plus the `end the file with true` escape hatch for a genuinely side-effect-only page script; line 88 no longer tells agents the exit code is meaningless.

## ⚠️ `rc-ending.mjs` will now start failing *correctly*

This is the one thing not to misread. `rc-ending.mjs --eval` computes `ok` honestly today, and its failures are currently invisible — a replay that never reaches the board prints `ok: false` and exits 0. After this lands, that same replay exits **3**.

**The first red exit from it is this fix working, not a regression this PR caused.** The same applies to `rc-room.mjs park --eval`. The passing path is verified below and still exits 0.

## How it was verified

The eight-case `about:blank` matrix from the plan. Probe files were written in the worktree and deleted before committing.

    # 1. ok:false                                        exit=3
    # 2. ok:true                                         exit=0
    # 3. throws                                          exit=3
    # 4. void 0                                          exit=3
    # 5. ({width: innerWidth}) -- no ok field            exit=0  (+ NOTE on stderr)
    # 6. plain screenshot, no eval file                  exit=0  (PNG written)
    # 7. failing run with an out path                    exit=3  (PNG written, 4740 bytes)
    # 8a. no arguments                                   exit=2
    # 8b. http://localhost:59999/                        exit=1

Case 1 stderr:

    shot.mjs: FAILED — the page script reported ok: false — deliberate
    {
      "ok": false,
      "error": "deliberate"
    }

Case 5 stderr:

    shot.mjs: NOTE the page script returned no "ok" field, so this run asserted nothing — exit 0 means Chrome ran and printed a value, not that anything passed.

**The real replay** — `npx next build`, `next start` on port 3150 (killed afterwards by the PID listening on that port, not by image name):

    node .agent/scripts/rc-ending.mjs orbited-in --eval shot150-ending.js      # exit 0
    node .agent/scripts/shot.mjs http://localhost:3150/games/race-condition 900 shot150-ending.png shot150-ending.js
    # exit=0 — ok:true, turnsPlayed 12, banner "Cyan wins", ringed [0,5,10,15]

**Negative control on that same real caller** — the identical generated page script pointed at a page with no board, so the assertion fails for a real reason rather than a synthetic one:

    node .agent/scripts/shot.mjs http://localhost:3150/ 900 - shot150-ending.js
    shot.mjs: FAILED — the page script reported ok: false — the mode picker never hydrated
    exit=3

**The rc-room fix, pinned old-shape against new-shape** through the gate, since the `park` half could not be run here (see below):

    ({ error: "board never appeared" })              -> exit=0  + NOTE   <- what it did before
    ({ ok: false, error: "board never appeared" })   -> exit=3           <- what it does now

That is the concrete demonstration that fixing `shot.mjs` alone would have left a real failure passing.

## Notes for the reviewer

- **`rc-room.mjs park --eval` could not be run end to end.** Room creation answered `500 server-error`; the server log shows Prisma rejecting `DATABASE_URL` in this worktree's provisioned env. The plan marked that half optional. The two edited lines are a pure shape change, and the old-vs-new probe above shows exactly what it buys — but nobody has driven a live room through the new exit path, so that is the weakest link in this verification. `rc-room.mjs check` needs the same database and was not run either.
- **`npx next build` needed `npx prisma generate` first** in a fresh worktree (`npm ci` did not run it). Unrelated to this change; noting it because the plan's verification block does not mention it.
- **Object detection is `typeof result === "object" && "ok" in result`.** An array, or a primitive like `42` or `"done"`, therefore lands in the no-`ok` measurement bucket and exits 0 with the NOTE. That is deliberate — a page script returning a bare number is a measurement — but it does mean a script returning the string `"failed"` passes. The convention is `ok`, and the NOTE is what makes the absence audible.
- **`payload.error` is set on both failure paths**, which adds an `error` key to stdout that was not there before. The plan called for this: `JSON.stringify` drops an `undefined` `result` key entirely, so a script that returned nothing would otherwise print a payload that said nothing at all. `{viewport, result}` is unchanged for every other run.
- The exit code for a falsy-but-not-false `ok` (`0`, `""`, `null`) is 3, and the banner prints the actual value via `JSON.stringify(result.ok)` so it does not read as `ok: false` when it was `ok: 0`.
- I did not add a flag to opt into gating. The plan rejected it explicitly — a gate behind a flag every future caller can forget is the same failure with an extra step.


csalinas-dev:
## 🤖 Review: APPROVED

Checked: security, correctness, efficiency, fit against the plan. Every exit path was
run on this machine against real headless Chrome, and **the `rc-room.mjs park`/`check`
gap the PR body called out has been closed** — I stood up a throwaway MySQL and ran
both end to end.

### Verified — the eight-case matrix, reproduced

`node .agent/scripts/shot.mjs about:blank 900 …` on Node v24.11.0, Chrome headless:

| case | expected | got |
|---|---|---|
| 1 `({ok:false, error:"deliberate"})` | 3 | **3** — payload on stdout, `shot.mjs: FAILED — the page script reported ok: false — deliberate` + the result object on stderr |
| 2 `({ok:true})` | 0 | **0**, silent |
| 3 throws | 3 | **3**, `payload.error` = `the page script threw: Error: boom …` — no unhandled-rejection stack |
| 4 `void 0;` | 3 | **3**, `the page script returned no value …` |
| 5 `({width: innerWidth})` | 0 | **0** + exactly one stderr NOTE |
| 6 plain screenshot, no eval file | 0 | **0**, PNG written, **no NOTE, no extra stderr** |
| 7 failing run with an out path | 3 | **3**, PNG written *before* the exit |
| 8a no args | 2 | **2** |
| 8b `http://localhost:59999/` | 1 | **1** |

`3, 0, 3, 3, 0, 0, 3, 2, 1` — matches the PR body exactly.

### Verified — shapes nobody enumerated

| page script returns | exit | note |
|---|---|---|
| `null` | **3** | payload carries both `"result": null` and the `error` |
| `Promise.resolve({ok:false,…})` | **3** | `awaitPromise` covers it |
| `Promise.reject(…)` | **3** | reported as a throw, payload has `error` |
| promise resolving after 8s | **3** | no premature timeout truncates the verdict |
| `({ok: 0})` | **3** | banner reads `reported ok: 0` |
| `({ok: false})` (no `error`) | **3** | banner omits the dangling em dash |
| `"just a string"`, `[]`, `false`, `true` | **0** + NOTE | non-object → measurement bucket, as the plan specifies |
| `({ok: "false"})`, `({ok: []})` | **0**, *silent* | truthy `ok`; see non-blocking #1 |

**Output integrity on the failing path.** A page script returning a 447 KB result
exits 3 with stdout intact and `JSON.parse`-able both when redirected to a file and
when piped through `cat` — `process.exit(3)` does not truncate the evidence. Case 7's
`fail.png` is a real 4740-byte PNG (`89504e47…` magic, `IEND` present), 900x1000, and
opens.

### Verified — the load-bearing `rc-room.mjs` pin

The exact old and new early-return shapes, through the new `shot.mjs`:

    (async () => ({ error: "board never appeared" }))()            -> exit 0 + NOTE
    (async () => ({ ok: false, error: "board never appeared" }))()  -> exit 3 + FAILED banner

So the two ok-less returns really would have been a real failure passing the gate, and
the `ok: false` added here really does close it. Confirmed live, not just synthetically:
pointing the *generated* `rc-room-*.js` at `/games/race-condition` (mode picker, no
board) drives the real early return and the command exits **3** with
`FAILED — the page script reported ok: false — board never appeared`.

All three return paths in the `park --eval` template now carry `ok` (lines 546, 551, 574);
`rc-ending.mjs`'s `fail()` helper and final return already did.

### Verified — `rc-room.mjs park` / `check` end to end (the gap in the PR body)

The implementor's blocker was a worktree with **no `node_modules`**, not a bad
`DATABASE_URL`. `npm ci` + a throwaway `mysql:8.0` on 127.0.0.1:13150 + `prisma db push`
+ `next build` + `next start -p 3150`:

- `node .agent/scripts/rc-room.mjs check --base http://localhost:3150` → **34 checks
  passed, 22 bodies scanned for tokens, exit 0**.
- `rc-room.mjs park --seats 1 --eval …` then `shot.mjs …?room=AUGM 900 rc-room.png <eval>`
  → `screenDisagreedWithTheRoom: 0, boardMovedOnItsOwn: 0, ok: true`, **exit 0**, PNG written.
- `rc-ending.mjs orbited-in --eval …` (exit 0) then its replay through `shot.mjs`
  → 12 turns played, endgame panel found, `ringed: [0,5,10,15]`, **exit 0**.

Both callers' *passing* paths still pass. Container and server torn down by PID/port.

### Acceptance criteria

- [x] `ok: false` → exit 3, result on stdout, FAILED banner + result on stderr — case 1
- [x] `ok: true` → exit 0 — case 2
- [x] throw → exit 3 with `error` naming the throw, no unhandled-rejection stack — case 3
- [x] no value → exit 3 — case 4
- [x] no `ok` field → exit 0 + exactly one NOTE — case 5
- [x] no eval file → exit 0, PNG, unchanged — case 6
- [x] failing run still writes the PNG — case 7, verified openable
- [x] usage 2 / dead server 1 unchanged — cases 8a/8b
- [x] every `return` in `rc-room.mjs`'s page-script template carries `ok`
- [x] `grep -rn "exits 0" .agent CLAUDE.md` surfaces only true statements (`shot.mjs:7`,
      `shot.mjs:226`, `CLAUDE.md:106` — all describing the no-`ok` measurement path)
- [x] `CLAUDE.md` documents exit 3, the NOTE, and the `true` escape hatch
- [x] Docs preserve the *record* of the silent green — `CLAUDE.md:88` and
      `rc-ending.mjs:237-242` still tell the mode-picker/`board not found`/zero-turns
      story; only the claim about the exit code changed, which is what was asked for

Scope is exactly the plan's four files (+58/-4 in `shot.mjs`, the rest comments and
docs). No probe files leaked into the branch (`git status` clean after the full run).
No security surface: no new input reaches a sink, no dependency added, and the only new
data on stderr is the page script's own already-printed result.

### Non-blocking observations

1. **A truthy non-boolean `ok` exits 0 with no stderr line at all.** `({ok: "false"})`
   and `({ok: []})` are neither a FAILED nor a NOTE — they pass silently, because
   `"ok" in result` puts them in the assertion bucket and `!result.ok` is false. This is
   exactly what the plan specifies (`failure` is set iff `!result.ok`), no committed
   caller produces the shape, and tightening it to `result.ok !== true` would be a
   different decision than the one that was approved. Worth knowing, not worth changing here.
2. **A falsy non-object result is a measurement, not a failure.** `false`, `0` and `""`
   land in the NOTE bucket and exit 0. Consistent with "ending with `true` is the
   side-effect-only escape hatch", but a page script that returns a bare `false` meaning
   "no" will read as green.
3. `readFileSync(evalFile, …)` sits inside the `try`, so an eval path that passes
   `existsSync` but cannot be read (a directory) is reported as `the page script threw:
   EISDIR` and exits 3 rather than 2. Cosmetic.
4. For the next agent: **`.agent/scripts/*` needs `npm ci` in the worktree first.**
   Without `node_modules`, `npx prisma` pulls 7.x from the registry and rejects this
   repo's v6 schema with `P1012` — which looks like a bad `DATABASE_URL` and is not one.

PR #160, opened 2026-08-16
docs: P1012 is not a database problem -- a fresh checkout needs npm ci

### LINKED ISSUE #159: docs(agents): a fresh worktree needs npm ci, and npx hides that with a misleading error
Three agents have lost verification coverage to this. The reviewer of #150 diagnosed it.

An agent worktree is a checkout with **no `node_modules`**. That part is ordinary. The expensive part is what happens next: `npx prisma` in a directory with no local install **downloads the latest Prisma (7.x)** instead of this repo's pinned v6, and 7.x rejects this schema with `P1012` — an error that reads as *"cannot reach the database"*.

So two agents reported "I could not get a database" and honestly scoped their verification down. Both were one `npm ci` away from a full run. Their disclosure is the only reason the pattern was visible at all.

### What would have prevented it

A line in `CLAUDE.md` under Development, next to the existing `npx prisma studio` / `npx prisma db push` commands, saying: run `npm ci` first in a fresh checkout, because `npx` will otherwise fetch a Prisma major this schema does not support and fail in a way that looks like a connection problem.

Worth checking whether the same trap applies to any other `npx` invocation the docs recommend.

### Done when
- `CLAUDE.md` warns that a fresh checkout needs `npm ci` before any `npx` tool invocation, and names the misleading `P1012` symptom specifically.
- Any other `npx`-based instruction in the repo's docs is checked for the same failure mode.

---

## 🤖 Implementation Plan

_Written by the architect agent on 2026-08-16 02:23 UTC. The implementor builds exactly this._

### Goal

`CLAUDE.md`'s Development section tells an agent in a fresh checkout to run `npm ci`
first, and gives it enough to recognise the `P1012` error it gets if it doesn't — so
that "the wrong Prisma CLI is running" stops being read as "there is no database".

### Approach

Documentation only. The warning goes in `CLAUDE.md` under **## Development**, directly
beneath the existing command fence, with `npm ci` added as the fence's first line. That
is the one place an agent already reads *before* running `npx prisma`. The alternative —
a new `## Troubleshooting` section — loses because it is only found after the agent is
already confused and has already scoped its verification down, which is exactly the
failure this issue is about.

Write it symptom → cause → fix, for a reader who is mid-misdiagnosis.

### Context the implementor needs — verified in a fresh worktree with no `node_modules`

Everything below was reproduced, not inferred. `package-lock.json` pins
`prisma` and `@prisma/client` at **6.19.3** (lines 7031 and 1773).

**1. `npx` fetches the wrong major.** In the worktree, `npx prisma --version` printed:

```
prisma               : 7.9.1
@prisma/client       : ^6.19.3
```

It resolves from the npx cache (`AppData/Local/npm-cache/_npx/...`), not the repo.

**2. `P1012` is emitted at `prisma/schema.prisma:13` for at least three unrelated
causes.** The error *code* and the *line it points at* are identical in all three; only
the `error:` text differs. All three were produced against this repo's schema:

| Cause | `error:` line | Fix |
|---|---|---|
| Prisma 7.x CLI (no `npm ci`) | ``The datasource property `url` is no longer supported in schema files. Move connection URLs for Migrate to `prisma.config.ts`...`` | `npm ci` |
| Prisma 6.19.3, `DATABASE_URL` unset | `Environment variable not found: DATABASE_URL.` | set `DATABASE_URL` |
| Prisma 6.19.3, `DATABASE_URL` not a MySQL URL | ``Error validating datasource `db`: the URL must start with the protocol `mysql://`.`` | fix the value |

The first is the trap. Its message names `url`, *connection* URLs and Accelerate, and
underlines the literal line `url = env("DATABASE_URL")` — which is why two agents read it
as "I have no database" and scoped their verification down. The schema is **not** wrong;
the CLI is. The disambiguator is the `Prisma CLI Version :` line the CLI prints at the
bottom of its own error: `7.9.1` means you skipped `npm ci`.

**3. Audit of every `npx` in the repo — six occurrences, two need this warning.**

- `CLAUDE.md:44` `npx prisma studio`, `CLAUDE.md:45` `npx prisma db push` — **exposed.**
  This is where the warning goes.
- `src/lib/realtime/README.md:205` `npx prisma db push` — **exposed.** Another doc telling
  a reader to run `npx prisma`; gets a one-line pointer, not a copy of the explanation.
- `Dockerfile:8` `RUN npx prisma generate` — safe, `npm ci` is line 6 of the same stage.
- `.github/workflows/deploy.yml:32` — runs inside that built image. Safe.
- `.agent/scripts/rc-ending.mjs:27` `npx next start` — a comment in a script whose
  documented flow already requires a built app (so `npm ci` has happened). Leave it.
- `package.json` scripts are all bare `next ...`, resolved from `node_modules/.bin`. With
  no install they fail with an unambiguous "not found" — no misleading-error risk.

### Changes

1. **`CLAUDE.md`**, `## Development` (lines 40–46) — add `npm ci` as the first line of the
   existing fence (`# install first -- see below`), and below the fence add a short prose
   warning in the file's existing voice. It must, in this order: state the symptom
   (`npx prisma` fails with `P1012` pointing at `schema.prisma:13`, reading like a
   database-connection problem); give the cause (no `node_modules` → `npx` downloads
   Prisma 7.x, which rejects `datasource.url`; the lockfile pins 6.19.3); give the fix
   (`npm ci`); and name the check (`Prisma CLI Version` at the foot of the error — if it
   is not 6.x you skipped the install). Explicitly separate the other `P1012`s: a
   *missing or non-`mysql://` `DATABASE_URL`* is a genuinely different failure with a
   different `error:` line, so read the message, not the code.
2. **`src/lib/realtime/README.md`**, "Local development" (~line 205) — one sentence after
   the `npx prisma db push` line: run `npm ci` first in a fresh checkout, or `npx` fetches
   a Prisma major this schema does not support; see `CLAUDE.md` → Development. Do not
   restate the whole explanation here.

### Out of scope

- Changing `prisma/schema.prisma` to satisfy Prisma 7, or bumping the Prisma dependency.
  The schema is correct for the pinned version.
- Adding npm scripts, an `engines` field, a `.npmrc`, or any tooling change. Docs only.
- `README.md` (human-facing, already says `npm install` and mentions no `npx`),
  `Dockerfile`, `deploy.yml`, `.agent/scripts/*` — audited above, all safe.

### Acceptance criteria

- [ ] `CLAUDE.md`'s Development fence starts with `npm ci`.
- [ ] The warning sits in `## Development`, immediately beside those commands — not in a
      new section elsewhere in the file.
- [ ] It names `P1012` verbatim, says the error looks like a connection problem, and says
      the actual cause is `npx` resolving Prisma 7.x instead of the pinned 6.19.3.
- [ ] It tells the reader to check the `Prisma CLI Version` line in the error output.
- [ ] It distinguishes that `P1012` from the `P1012` caused by an unset or non-`mysql://`
      `DATABASE_URL`, so the two are not conflated.
- [ ] `src/lib/realtime/README.md` carries a one-line `npm ci` pointer next to its
      `npx prisma db push` instruction.
- [ ] No non-documentation file is modified.

### Verification

    git diff --stat        # CLAUDE.md and src/lib/realtime/README.md only

No build or test run is needed or expected — this PR changes no shipped code. If you want
to see the trap for yourself, in a checkout with no `node_modules`:

    npx --yes prisma --version     # prints 7.9.1 next to "@prisma/client : ^6.19.3"
    npx --yes prisma validate      # P1012, "datasource property `url` is no longer supported"

### Risks

Low. The only real failure mode is writing the warning so it reads as "Prisma is broken"
or "you need a database" — the exact misreading it exists to prevent. Lead with the
symptom the confused agent is staring at, then the cause, then `npm ci`.

csalinas-dev:
Closes #159

Documentation only — no shipped code changes.

## What changed

- **`CLAUDE.md`, `## Development`** — `npm ci` is now the first line of the command fence (`# install first -- a fresh checkout has no node_modules; see below`), followed by the warning the issue asks for, written symptom → cause → fix for a reader who is already mid-misdiagnosis: `npx prisma` with no `node_modules` fails with **`P1012`** underlining `url = env("DATABASE_URL")` at `prisma/schema.prisma:13`, which reads as a connection failure; the actual cause is `npx` having no pinned CLI to resolve and downloading Prisma 7.x instead of the 6.19.3 in `package-lock.json`.
- The warning **does not** say "`P1012` means you forgot `npm ci`." That would be a third wrong diagnosis wearing the authority of documentation. `P1012` at `schema.prisma:13` has at least three unrelated causes — same code, same underlined line, three different `error:` texts — so the doc gives them as a table and tells the reader to diagnose from the **`error:` text** and the **`Prisma CLI Version :`** line at the foot of the output, not from the code or the line number. It says explicitly that `npm ci` fixes neither of the other two.
- **`src/lib/realtime/README.md`, "Local development"** — one sentence beside its `npx prisma db push` line pointing at `CLAUDE.md` → Development. Deliberately not a second copy of the explanation.

Per the architect's audit, the other four `npx` occurrences (`Dockerfile:8`, `.github/workflows/deploy.yml:32`, `.agent/scripts/rc-ending.mjs:27`, and `package.json`'s bare `next` scripts) are untouched — each either runs after `npm ci` in the same stage, runs inside a built image, or fails unambiguously.

## How it was verified

All three `P1012` variants reproduced in this worktree, against this repo's real `prisma/schema.prisma`. Every `error:` string in the table is copied from output below, not paraphrased.

**1. Prisma 7 via `npx`, no `node_modules`** — the trap:

    $ npx --yes prisma --version
    prisma               : 7.9.1
    @prisma/client       : ^6.19.3
    Schema Engine        : ... (at ..\AppData\Local\npm-cache\_npx\...)

    $ npx --yes prisma validate
    Error code: P1012
    error: The datasource property `url` is no longer supported in schema files. ...
      -->  prisma\schema.prisma:13
    13 |   url      = env("DATABASE_URL")
    Prisma CLI Version : 7.9.1

**2. `npm ci`, then the same command** — the pinned CLI, this worktree's provisioned `.env`:

    $ npm ci && npx prisma validate
    Error code: P1012
    error: Error validating datasource `db`: the URL must start with the protocol `mysql://`.
      -->  prisma\schema.prisma:13
    Prisma CLI Version : 6.19.3

**3. Pinned CLI, no `DATABASE_URL`** (schema copied to an empty scratch dir so no `.env` is found):

    $ .../node_modules/.bin/prisma validate
    Error code: P1012
    error: Environment variable not found: DATABASE_URL.
      -->  prisma\schema.prisma:13
    Prisma CLI Version : 6.19.3

Diff scope:

    $ git diff --stat
     CLAUDE.md                  | 21 +++++++++++++++++++++
     src/lib/realtime/README.md |  8 +++++---

Self-check the issue demands: staring at output 1, the new text names `P1012`, says it looks like a connection problem, and sends me to `Prisma CLI Version : 7.9.1` → `npm ci`. Doing that lands on output 2, which the same table identifies as a *different* failure `npm ci` will not fix. Both hops are ones the doc actually makes.

## Notes for the reviewer

- **Output 2 is worth dwelling on.** It is what an agent gets in a correctly provisioned worktree that *did* run `npm ci`. If the doc had reduced to "`P1012` → run `npm ci`", the very next thing the reader hits after following it is another `P1012` the advice does not explain — which is why the `error:`-text-not-code framing is the load-bearing part of this change and not hedging.
- The table is a mild style deviation: `CLAUDE.md` is otherwise fences and prose bullets, no tables. I used one because "identical code, identical line, three different messages" is the point, and a table shows that structurally in a way three sentences don't. Easy to flatten to bullets if you'd rather.
- The Prisma 7 `error:` text is quoted only as far as its first sentence; the real message continues into `prisma.config.ts` / Accelerate advice. The first sentence is enough to match on and the full one would not fit a table cell.
- `7.9.1` is what `npx` resolved today. The doc says "the latest Prisma (7.x)" rather than pinning that number so it doesn't rot; the exact version only appears here.
- No build or test run: this PR changes no shipped code, and the plan's Verification section is `git diff --stat`. `npm ci` was run anyway, purely to reproduce variants 2 and 3.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -40,10 +40,31 @@ GAME_ROOM_RATE_LIMIT        # optional; room creations/minute/IP for POST /api/r
 ## Development
 
 ```bash
+npm ci           # install first -- a fresh checkout has no node_modules; see below
 npm run dev      # start dev server on localhost:3000
 npx prisma studio  # inspect DB
 npx prisma db push # apply schema changes
 ```
+
+`npx prisma` in a checkout with no `node_modules` fails with **`P1012`** underlining
+`url = env("DATABASE_URL")` at `prisma/schema.prisma:13`, which reads like "I can't reach
+the database". It isn't. With nothing installed locally, `npx` has no pinned CLI to
+resolve and downloads the latest Prisma (7.x) instead of the 6.19.3 in
+`package-lock.json`; 7.x no longer accepts `url` inside a `datasource` block. The schema
+is correct — the CLI is the wrong major. Run `npm ci`.
+
+Diagnose from the `error:` text and the `Prisma CLI Version :` line at the foot of the
+output, **not** from the code or the line number: `P1012` at `schema.prisma:13` has at
+least three unrelated causes, all of which this repo produces.
+
+| `error:` says | Cause | Fix |
+|---|---|---|
+| ``The datasource property `url` is no longer supported in schema files.`` | `npx` fetched Prisma 7.x — the `Prisma CLI Version` line will say `7.x`, not `6.x` | `npm ci` |
+| `Environment variable not found: DATABASE_URL.` | `DATABASE_URL` is unset | set it |
+| ``Error validating datasource `db`: the URL must start with the protocol `mysql://`.`` | `DATABASE_URL` is set to something that isn't a MySQL URL | fix the value |
+
+The last two are genuine environment problems and `npm ci` fixes neither, so read the
+message before reaching for a fix.
 
 ## Important Patterns
 
--- a/src/lib/realtime/README.md
+++ b/src/lib/realtime/README.md
@@ -202,6 +202,8 @@ await send({ type: "PLACE", cell: 4 }, { optimistic: true });
 ## Local development
 
 There is no MySQL in the repo's dev setup by default; `docker compose up db`
-provides one. Apply the schema with `npx prisma db push`. In production the
-deploy workflow runs `prisma db push --skip-generate` against the freshly built
-image *before* the app serves traffic.
+provides one. Apply the schema with `npx prisma db push` — run `npm ci` first in a
+fresh checkout, or `npx` fetches a Prisma major this schema does not support and fails
+with a `P1012` that looks like a connection error (`CLAUDE.md` → Development). In
+production the deploy workflow runs `prisma db push --skip-generate` against the
+freshly built image *before* the app serves traffic.


csalinas-dev:
## 🤖 Review: APPROVED

Docs-only, reviewed proportionately but with the one failure mode that matters for docs checked properly: **is it confidently wrong?** It is not. All three `P1012` variants were re-reproduced independently in a fresh reviewer worktree against this repo's real `prisma/schema.prisma`, and every `error:` string in the new table matches real CLI output character-for-character.

**Verified (reviewer worktree `rev-159`, no `node_modules`, orchestrator-provisioned `.env`)**

1. `npx --yes prisma --version` -> `prisma : 7.9.1` next to `@prisma/client : ^6.19.3`, engines resolved from `AppData/Local/npm-cache/_npx/...`. The trap is real and current.
2. `npx --yes prisma validate` -> `Error code: P1012`, ``error: The datasource property `url` is no longer supported in schema files. ...``, `--> prisma\schema.prisma:13` underlining `url = env("DATABASE_URL")`, `Prisma CLI Version : 7.9.1`. Matches row 1 verbatim (the doc quotes the first sentence; the truncation is disclosed in the PR body and the sentence boundary is real, not a mangled quote).
3. `npm ci` then `npx prisma validate` -> ``error: Error validating datasource `db`: the URL must start with the protocol `mysql://`.``, `Prisma CLI Version : 6.19.3`. Matches row 3 verbatim.
4. Schema copied to an empty scratch dir (no `.env`), run through `node_modules/.bin/prisma` -> `error: Environment variable not found: DATABASE_URL.`, `6.19.3`. Matches row 2 verbatim.
5. **Extra check the plan did not do — is row 1 conditional on `DATABASE_URL`?** Ran `npx --yes prisma@7.9.1 validate` in the no-`.env` scratch dir: still the ``datasource property `url``` message, not `Environment variable not found`. Prisma 7 rejects at schema-parse, before env resolution. So the table is a genuine partition — row 1's `error:` text implies "wrong CLI major" unconditionally, and the doc's "diagnose from the `error:` text" instruction is sound rather than merely usually-true.
6. Pins: `package-lock.json` has `prisma`/`@prisma/client` at `6.19.3` (`package.json` carries `^6.19.3`). The doc correctly points at the *lockfile*, since the caret range is what makes "6.19.3" the right number to cite.
7. Untouched `npx` sites spot-checked against the architect's audit, and the audit holds: `Dockerfile:8` is preceded by `npm ci` on line 6 of the same stage, and the runner stage is `COPY --from=builder /app ./`, so `deploy.yml:32`'s `npx prisma db push` inside the image resolves the vendored 6.19.3 rather than downloading. `.agent/scripts/rc-ending.mjs:27` is a comment in a flow that already requires a built app.
8. `git diff --stat origin/main...HEAD` -> `CLAUDE.md` and `src/lib/realtime/README.md` only, +26/-3. `git grep -n "npx "` shows the other four occurrences unchanged. Nothing swept in. `mergeStateStatus: CLEAN`.

**Would it have worked on me?** Yes, first hop. Reading the raw Prisma 7 output cold, the new text's *first sentence* names `P1012`, the underlined `url = env("DATABASE_URL")`, `schema.prisma:13`, and the "reads like I can't reach the database" misread — it matches on everything I would be staring at — and ends "Run `npm ci`." Following that lands on variant 3, which the same table identifies as a different failure `npm ci` will not fix, so the doc does not strand the reader on the second hop either.

**Does it anywhere imply "`P1012` means you forgot `npm ci`"?** No, and I looked for it specifically. Both statements run in the safe direction (no `node_modules` -> this `P1012`), never the converse, and `CLAUDE.md:66` says outright that `npm ci` fixes neither of the other two. That is the part that would have made the doc worse than nothing, and it is handled.

**Acceptance criteria**
- [x] Development fence starts with `npm ci` — `CLAUDE.md:43`
- [x] Warning sits in `## Development` beside the commands, not a new section — `CLAUDE.md:49-67`, immediately under the fence
- [x] Names `P1012` verbatim, says it looks like a connection problem, names `npx` resolving 7.x instead of 6.19.3 — `CLAUDE.md:49-54`
- [x] Tells the reader to check `Prisma CLI Version` — `CLAUDE.md:56` and table row 1
- [x] Distinguishes the unset / non-`mysql://` `DATABASE_URL` variants — table rows 2-3 plus `CLAUDE.md:66`
- [x] `src/lib/realtime/README.md` carries a one-line pointer, not a second copy — `README.md:205-208`
- [x] No non-documentation file modified

**On the flagged style deviation — keep the table.** Calling it: this is the right exception. `CLAUDE.md` is otherwise fences and prose bullets, but the whole point here is "identical code, identical line number, three different messages", and a three-row table is the only form that shows that as a partition instead of asking the reader to hold three near-identical sentences in their head. It also survives the medium: `CLAUDE.md` is consumed as raw text by an agent, where a pipe table is still an obvious lookup structure. Not worth flattening.

**Non-blocking observations**
- In an orchestrator-provisioned worktree, `DATABASE_URL` is set to something that is not a `mysql://` URL, so an agent who follows this doc correctly (`npm ci`, rerun) lands on row 3, whose Fix cell reads "fix the value" — not actionable for an agent that did not write that `.env` and should not edit it. A future touch-up could make row 3's fix "point it at a MySQL instance — `docker compose up db`", which is what `src/lib/realtime/README.md:204` already says. Out of scope here; the row is factually correct as written.
- Read standalone, the `src/lib/realtime/README.md` sentence is the one place a hurried reader could compress to "`P1012` -> `npm ci`". The `CLAUDE.md` -> Development cross-reference is in the same sentence and duplicating the table would be worse, so leave it.
- Path separators in the doc are `prisma/schema.prisma:13`; real Windows output prints `prisma\schema.prisma:13`. Repo convention, and it does not impede matching.

(Posted as a comment rather than a review approval: GitHub refuses `--approve` on a PR authored by the same account.)

PR #168, opened 2026-08-16
feat(blog): sync Substack on the read path, behind a 10-minute single-flight gate

csalinas-dev:
Closes #154

Built the re-planned design (sync-on-read behind a throttle, plus deploy-time sync). No cron, no HTTP endpoint.

## What changed

**The trigger — `src/lib/substack/refresh.js`.** `scheduleSubstackRefresh()` does exactly one thing: `after(() => gate()())`. No await, no import, no work of any kind in the calling frame. Both blog routes call it as their first statement — `/blog`, and *both* `generateMetadata` and `Page` in `/blog/[slug]`, since `generateMetadata` can `notFound()` before `Page` runs. `REFRESH_INTERVAL_MS = 10 min` and `MAX_RUN_MS = 2 min` live here, with the reasoning in comments.

**The throttle — `src/lib/substack/throttle.js`.** A pure `createRefreshGate({ run, intervalMs, maxRunMs, now })`, zero imports so plain `node` can load it. Its header carries the three reasons it is **not** `unstable_cache` (it bounds a *write*; `unstable_cache` gives no single-flight across concurrent misses; an eviction would silently un-throttle a write path). Pinned to `globalThis` like `src/lib/prisma.js` so two route entries cannot get two gates — verified below by interleaving both routes.

**The single-flight invariant is commented at the site.** Everything from the in-flight check to the assignment of `inFlight` runs in one synchronous turn; an `await` between them reintroduces the stampede and nothing fails visibly. I proved the offline check catches exactly that (RED output below).

**`src/lib/substack/run.js`** — the one place a sync is performed, shared by the read path and the CLI. It calls `getFeedUrl()` inside its own try (it is a *default parameter* of `syncSubstackPosts`, so an unset `SUBSTACK_FEED_URL` throws before any report exists), imports `./sync` dynamically to keep the ingestion graph off the render path, and never throws.

**Failure surfacing.** `SubstackSyncStatus` (one row, `id: 1`, purely additive) is the durable record; `src/lib/substack/status.js` can never throw; `src/lib/substack/alert.js` mails `SUBSTACK_ALERT_EMAIL` after ≥2 failures spanning ≥30 min, at most once per 24h, with a recovery mail, using the SMTP transport `src/lib/auth.js` already configures.

[...]

Issue #166, opened 2026-08-16
docs(agents): the documented way to seed a local DB for the blog gates cannot be run on this machine


csalinas-dev:
The repo's only documented way to seed a local database for the blog verification gates cannot be run by an agent on this machine.

Two places document it:

- `.agent/scripts/verify-blog-overflow.mjs:11`
- `.agent/fixtures/substack/feed-overflow.xml:32`

Both instruct the reader to invoke the seeding script with Node's `--env-file` flag pointed at the worktree's local environment file. On this machine that command is refused by the environment-file safety hook before it runs.

### Why this is worth fixing rather than working around

An implementor on #164 followed the documented instruction, was refused, and correctly stopped rather than rephrasing to get past a safety check. That cost a full review round — the gate went unrun and the issue was blocked. The reviewer then ran the same gate successfully by a *different* route: starting a throwaway MySQL container and passing the connection string in the command's own environment, never referencing the local environment file at all.

So the capability exists and works. Only the documented recipe is unusable, which is the worst combination: an agent that follows the docs is blocked, and an agent that ignores them succeeds.

### What the fix looks like

Replace the `--env-file` recipe in both places with the route that is known to work on this machine — a throwaway container plus the connection string passed inline to the command that needs it. That is also the pattern the repo's own guidance already recommends for handing a credential to a single command.

While editing, check whether any other `.agent/` script or fixture documents the same recipe.

### Out of scope

The safety hook's own behaviour. It is machine configuration rather than repository content, and it is being raised with Christopher separately — see the comment thread on #164. This issue is only about the repository's documented instructions being executable.

### Done when
- Neither documented seeding recipe depends on a form that is refused on this machine.
- The replacement has been run end to end, and the gate it seeds passes.
- No other `.agent/` file documents the unusable form.

---

## 🤖 Implementation Plan

_Written by the architect agent on 2026-08-16 07:55 UTC. The implementor builds exactly this._

### Goal

Every place in the repo that tells an agent how to seed a local database for the blog
gates gives one recipe, and that recipe is a throwaway MySQL container plus a connection
string passed in the invoking command's own environment. No documented recipe points a
command at a checked-out environment file any more.

Branch base is **`epic/substack-rss`**, not `main` — all six files below live there.

### Approach

**One canonical recipe, five pointers.** The full container-and-teardown procedure goes in
the header of `.agent/scripts/seed-substack-fixture.mjs`, because that is the script being
[...]

Issue #164, opened 2026-08-16
chore(epic): catch epic/substack-rss up with main before part 3


csalinas-dev:
`epic/substack-rss` is **2 ahead, 8 behind** `main`, and a `git merge main` conflicts. It must be caught up before part 3 (#154) is built on it, and certainly before the epic's integration PR.

### Why it was allowed to drift

Deliberately. #153's acceptance gate compared the branch against an unmodified copy of *itself* to prove existing MDX posts render identically; merging `main` mid-flight would have buried that diff in unrelated changes. That reason expired when #153 merged — this is the catch-up it was deferred for, not an oversight.

### What is on `main` that this branch lacks

- #148 — the Race Condition epic (a whole game, plus `/games`, nav and policy-page edits)
- #157 — the realtime seat-lockout fix, touching `src/lib/realtime/`
- #158 — `shot.mjs` now exits non-zero when a page script reports failure
- #160 — the `npm ci` / `npx` / `P1012` documentation
- #163 — the privacy-policy correction

### The conflicts that need a judgement, not just a resolution

- **`CLAUDE.md`** — both sides have added sections. This is a merge of content, not a pick-a-side.
- **`.agent/scripts/shot.mjs`** — `main` has #158's exit-code behaviour; this branch does not. `main` wins, but see the consequence below.
- **`src/app/(pages)/privacy-policy/page.js`** — #163 corrected it on `main`; #154's plan intends to edit the same paragraphs later. Both sets of facts must survive.

### The consequence that must not be missed

Once `main` is merged, `shot.mjs` on this branch **gains** #158's exit-code behaviour. `verify-blog-dom.mjs` and `verify-blog-overflow.mjs` were written with a workaround — they take their exit code from their own runner because `shot.mjs` could not fail. That workaround becomes *redundant*, not *wrong*. Do not delete it as cleanup without re-proving each gate can still go red; a gate that silently stops failing is the exact defect #158 existed to fix.

### Done when
- `epic/substack-rss` contains `main`, with every conflict resolved so that both sides' facts survive.
- `verify-blog-dom.mjs`, `verify-blog-overflow.mjs`, `verify-blog-posts.mjs` and `verify-substack-sync.mjs` all pass on the merged branch — and each has been shown to still go **red** when its subject is broken, since the merge changed the tooling underneath them.
- The DOM baseline is re-established against the merged branch, so #154 has a valid comparison point.
- `npm run lint` and `npm run build` pass.

---

[...]

Issue #162, opened 2026-08-16
ops: SUBSTACK_FEED_URL must be set in production before the blog sync can run


csalinas-dev:
Raised by the architect of #154 (part 3 of epic #151), which flagged it rather than routing around it.

`SUBSTACK_FEED_URL` is not present in the production environment file on the server, and **no PR can put it there** — that file is not in the repository, by design.

Until it is set, the scheduled sync will exit `3` ("could not start") and file its own alert issue. That is the design working correctly, not a fault, but it means the feature is inert on arrival.

### What is needed

- The publication's official RSS feed URL — the `/feed` endpoint of your Substack publication — set as `SUBSTACK_FEED_URL` on the server, around the time epic #151 merges.

### Two things worth deciding at the same time

1. **Your publication currently has zero posts.** The first real sync will therefore hit a single-item or empty feed. That path is covered by a dedicated fixture (it is the case where the XML parser returns an object rather than an array), but it does mean the feature cannot be observed doing anything useful until you publish something.
2. **The sync is planned to run every 30 minutes and on every successful deploy.** If you would rather it ran less often, say so before #154 merges — changing a cron line later is trivial, but the cadence is worth being deliberate about, since each run fetches a third party.

No agent work is blocked on this: #154 can be built, reviewed and merged without it. Only the feature actually functioning in production depends on it.

csalinas-dev:
I don't want it on a schedule sync. I think it should sync on page load and cache the data on the server for a set time period. On deployment is fine too. 

I'll write posts later.

Where do I get the RSS feed URL?

csalinas-dev:
**Where the feed URL comes from:** every Substack publication serves RSS at its own address with `/feed` appended.

- On a `substack.com` subdomain: `https://<publication>.substack.com/feed`
- On a custom domain: `https://<yourdomain>/feed`

Open it in a browser to confirm — you should get XML starting with `<rss ...>` and a `<channel>` containing your publication's title. That is the whole value; there is no key or token. If you paste the URL here (or just the publication name) I will put it in the issue so the implementor can point a fixture at the real shape.

Note it works even with zero posts — you get a valid feed with no `<item>` elements, which is a case the ingestion code already handles.

---

**On the sync trigger — you are overriding your own spec, and I want that on the record rather than assumed.**

[...]

PR #169, opened 2026-08-16
epic(blog): Substack as the writing surface, the site as the reading surface

### LINKED ISSUE #151: Integrate Substack as native blog content source via RSS
**Epic branch: `epic/substack-rss`** — every sub-issue PR targets this branch, not `main`. When all parts have merged, a final PR merges `epic/substack-rss` into `main`.

---

## Summary

Use Substack as the authoring/newsletter platform while continuing to render blog posts natively on `csalinas-dev-site` with the site's existing components and styling.

Published Substack posts should be ingested from the publication's official RSS feed, normalized into the site's blog model, stored locally, and served from the existing `/blog` and `/blog/[slug]` routes. Do not embed Substack pages or rely on undocumented Substack APIs.

## Current implementation

The blog currently uses build-time MDX posts registered manually in `src/content/posts/index.js`. Each post exports metadata plus an MDX component, and `src/app/(pages)/blog/[slug]/page.jsx` renders the post through the site's existing native article layout.

The app already uses Prisma with MySQL, so Substack content can be persisted locally rather than requiring a live Substack request on every page view.

## Desired architecture

```text
Substack
   ↓ publish
Official RSS feed
   ↓ sync
Normalize + sanitize
   ↓
Prisma / MySQL
   ↓
Existing Next.js blog routes
   ↓
Native site components/styles
```

Substack should only be responsible for authoring, email distribution, and its own subscriber ecosystem. The website should own presentation, URLs, metadata, navigation, analytics, and styling.

### 1. Substack feed configuration

- Add a configuration value such as `SUBSTACK_FEED_URL`.
- Read the publication's official `/feed` RSS endpoint server-side.
- Add an RSS parser dependency if needed.
- Keep all Substack-specific parsing isolated behind a small service/module so the rest of the blog does not depend directly on RSS field names.

### 2. Normalize posts

Map each RSS item into a stable internal shape, for example:

```js
{
  sourceId,
  slug,
  title,
  description,
  contentHtml,
  publishedAt,
  coverImage,
  sourceUrl,
}
```

Requirements:

- Use a stable Substack identifier/GUID when available rather than the title as identity.
- Derive a deterministic local slug.
- Preserve the canonical Substack URL as `sourceUrl` for attribution/debugging.
- Extract the post description/excerpt and hero/cover image when available.
- Handle edited posts so a subsequent sync updates the local copy.

### 3. Sanitize imported HTML

Substack article content should never be rendered directly without sanitization.

- Sanitize imported HTML server-side before persistence or rendering.
- Allow the markup required for normal article content such as headings, paragraphs, links, lists, blockquotes, code/pre, images, emphasis, and horizontal rules.
- Remove scripts, event handlers, iframes/embeds unless explicitly supported, and other unsafe markup.
- Normalize external links as appropriate (`rel`, target behavior, etc.).

### 4. Persist synced posts with Prisma

Add a model for externally sourced blog posts. Exact naming is flexible, but it should support at least:

- unique source ID / GUID
- unique local slug
- title
- description/excerpt
- sanitized HTML content
- cover image URL
- published timestamp
- original Substack URL
- last synced timestamp
- optional source/content hash for efficient update detection

Use `@db.LongText` for article HTML if necessary with the existing MySQL datasource.

### 5. Add an idempotent sync path

Create a reusable server-side sync function that:

1. Fetches the configured RSS feed.
2. Parses and normalizes entries.
3. Sanitizes article HTML.
4. Inserts new posts.
5. Updates changed posts.
6. Leaves unchanged posts alone.
7. Reports useful sync results/errors.

The sync function should be callable from a deployment/cron mechanism without duplicating ingestion logic. Do not make normal blog page requests responsible for synchronizing the feed.

### 6. Integrate with the existing blog

Refactor the current blog data layer so the UI consumes a common post shape regardless of source.

The existing native layout should remain the presentation layer, including:

- `/blog`
- `/blog/[slug]`
- article title/description
- publish date
- reading time
- hero image
- previous/next navigation
- SEO/Open Graph metadata
- existing MUI/styled components

Substack HTML should render inside the existing `Article` / `PostBody` layout rather than inheriting Substack styling.

### 7. Preserve existing MDX posts

Existing local MDX posts should continue to work unless they are intentionally migrated later.

Create a unified repository/data layer that can return both local MDX and synced Substack posts. Define deterministic behavior for slug collisions rather than silently overwriting one source with the other.

### 8. Reading time

Calculate reading time for imported Substack content so it continues to populate the existing metadata rail. The value should be based on the normalized article text rather than copied presentation markup.

### 9. Caching and failure behavior

- Blog pages should read from the local database and continue to work if Substack is temporarily unavailable.
- A failed feed sync must not remove or invalidate the last successfully synced posts.
- Log enough information to diagnose feed/parsing failures without logging unnecessary article content.

## Acceptance criteria

- [ ] Publishing a public post on Substack can result in a corresponding native post on `csalinas-dev-site` after the sync runs.
- [ ] No Substack iframe/embed is used for article rendering.
- [ ] Website article typography/layout is controlled entirely by the site's existing components/styles.
- [ ] Imported HTML is sanitized before being rendered.
- [ ] Synced posts are stored locally with Prisma/MySQL.
- [ ] Re-running the sync is idempotent and updates edited Substack posts rather than creating duplicates.
- [ ] Existing MDX posts continue to render correctly.
- [ ] `/blog` includes both existing local posts and synced Substack posts in the correct date order.
- [ ] `/blog/[slug]` supports synced posts with metadata, hero image, reading time, and previous/next navigation.
- [ ] Blog pages remain available when Substack or its RSS feed is temporarily unavailable.
- [ ] The implementation uses the official RSS feed and does not rely on undocumented Substack API endpoints.

## Non-goals

- Recreating Substack's editor in this application.
- Embedding Substack article pages.
- Syncing Substack comments, likes, Notes, or subscriber data.
- Accessing paid/private posts that are not present in the public RSS feed.
- Publishing from this website back to Substack.
- Depending on reverse-engineered/internal Substack API endpoints.

## Future considerations

Once basic ingestion is working, possible follow-ups include displaying a native subscribe CTA, tracking the original Substack URL, feed-specific categories/tags, and triggering sync automatically from the site's deployment/cron infrastructure.

csalinas-dev:
Integration PR for epic #151. Merges `epic/substack-rss` into `main`.

Closes #151.

### What this is

Substack becomes the place you write; the site stays the place people read. Published posts are pulled from the publication's own RSS feed, sanitized, stored in the database, and rendered through the existing `Article` layout at `/blog` and `/blog/[slug]` — same typography, same metadata, same navigation as the MDX posts, which keep working untouched.

### The three parts

| | | |
|---|---|---|
| #152 | ingestion — schema, feed parsing, sanitization, idempotent sync | PR #156 |
| #153 | unified data layer, native rendering, reading time, prev/next, SEO | PR #161 |
| #154 | sync trigger, throttle, and operational reporting | PR #168 |

Plus #164 (caught the branch up with `main`, so this is a clean merge rather than an archaeology exercise) and #166 (the documented way to seed a local database for the gates, which did not execute).

### The design decision you should know about

**The sync runs on page load, behind a 10-minute server-side throttle, plus once per deploy.** #151 originally said the opposite — that page requests must not be responsible for syncing. Christopher overrode that deliberately (#162), and the trade-offs were put in front of him before he did: a visitor's request now pays for the outbound fetch on a cold cache, concurrent cold requests could stampede, and failure gets quieter than a scheduled job's would.

Each of those has an answer in the code rather than a hope:

- The page responds first; the sync runs after the response, so no visitor waits on Substack.
- One throttle gate, pinned across module reloads, gives ten simultaneous requests a single shared sync. Verified with twenty interleaved requests across both routes.
- Failure surfaces two ways with complementary blind spots — email after repeated failures spanning at least half an hour, and an auto-filed issue on deploy — plus a status row so a failure outlives the process that saw it.

### Third-party HTML is the security boundary

The sanitizer took three review rounds. The allowlist is data, not hand-written rules, and an audit throws at import time if a URL-bearing attribute is ever declared without a validator — so the class of bug is closed, not just the instance. Every iframe becomes a labelled link rather than a live third-party frame. A reviewer wrote its own 80-payload battery, confirmed each of the library's 2026 CVEs is blocked at the pinned version, and could not get script into the output.

Two real defects were caught before merge and both were latent, not theoretical: a relative URL surviving on `<source srcset>` inside `<picture>` (which would have made a visitor's browser fetch a path a third party chose), and a URL validator that split on a character Substack's CDN puts *inside* its URLs — which stripped every responsive image from every post while reporting that nothing was dropped.

### What was verified, and how

There is no CI here, so the checking is scripts and browsers:

- `verify-substack-sync.mjs` — sync behaviour offline, no network, no database
- `verify-blog-posts.mjs`, `verify-blog-dom.mjs` — existing MDX posts render **identically**, proven against a baseline captured from a clean worktree, with the gate itself proven able to fail
- `verify-blog-overflow.mjs` — no page overflows its viewport at 430px or 1200px, on nine pages including deliberately hostile content
- `verify-substack-refresh.mjs` — the throttle, including that ten concurrent callers produce one sync

Failure modes exercised against a live server: unreachable feed, hanging feed, slow dribble, HTTP 500, HTML instead of XML, malformed XML, empty feed, dead database — and a dead database *and* a hanging feed together, which still serves `/blog` from the MDX posts.

### Known and deliberate

- The publication has no posts yet, so the first real sync will hit a single-item or empty feed. Both paths have fixtures; the single-item case is the one where the XML parser returns an object rather than an array.
- `SUBSTACK_ALERT_EMAIL` is unset, so the email channel is dormant until an operator sets it. The deploy channel works regardless.
- One gate's failure message cannot distinguish "no rows" from "no database", because it never sees an HTTP status. Documented where it bites; a follow-up is filed separately.


csalinas-dev:
## 🤖 Review: APPROVED

Integration review of the whole branch as one change landing on `main`. I did not re-run the parts' reviews; I built the merged branch, served it against a real MySQL, and played the blog.

### 1. Integration — an MDX post and a synced post adjacent

Seeded `feed-base.xml` (three 2026 posts) plus a hand-written fixture with posts dated **2024-05-01** (between `goldwater-bank` 2024-07-22 and `hashtag` 2024-02-27) and **2024-01-01** (older than everything), so the seam is crossed in both directions and both ends of the list are exercised.

`/blog` order — correct, interleaved by date across both sources:

```
first-post, 2026, no-enclosure, goldwater-bank, sandwiched, hashtag, wordleverse, oldest-of-all
```

Walking `/blog/<slug>` for all eight, the prev/next chain is a single unbroken line, and every crossing works:

| post | ← Previous | Next → |
|---|---|---|
| first-post (synced, newest) | 2026 | — |
| no-enclosure (synced) | **goldwater-bank (mdx)** | 2026 |
| goldwater-bank (mdx) | **sandwiched (synced)** | **no-enclosure (synced)** |
| sandwiched (synced) | **hashtag (mdx)** | **goldwater-bank (mdx)** |
| wordleverse (mdx) | **oldest-of-all (synced)** | hashtag |
| oldest-of-all (synced, oldest) | — | **wordleverse (mdx)** |

Both ends render exactly one footer link. A synced post carries category absent (not a placeholder), a real date (`May 1st, 2024`), a reading time (`1 min read`), a remote `` hero, `Read on Substack` in the header's external-link slot, `<h2>` ids including de-duplication (`a-second-heading`, `a-second-heading-2`), and `og:title` / `og:description` / `og:image` / `og:type=article` in the same shape as an MDX post with no canonical pointing at substack.com. An `<h1>` in feed content is demoted to `<h2>` and gets a TOC id. No `<iframe>` reaches the page.

### 2. Blast radius on `main`

`git diff --stat origin/main...HEAD` touches **nothing** under `src/lib/realtime/`, `src/app/(pages)/games/`, or `src/app/api/`. Outside the blog and the two new lib directories the change is: `prisma/schema.prisma` (+55/-0, two new models only), `.github/workflows/deploy.yml` (+99/-0), `package.json` (3 deps + 1 script), `CLAUDE.md`, `privacy-policy/page.js` (prose), and `src/lib/mdx/remark.mjs` + `src/mdx-components.js`, each extracting a constant/function to the shared module verbatim.

Runtime confirmation on the merged build — all 200, and a room created on each of the four registered games:

```
/games  /games/tic-tac-overflow  /games/edge-case  /games/connect-404
/games/race-condition  /games/hashtag  /games/wordleverse  /github  /   -> 200

POST /api/rooms   tto -> RJZ2    edge-case -> UQD5
                  connect-404 -> JYVZ    race-condition -> PCH6
                  (all slot=0, status=lobby, players=1)
```

Also confirmed `getPost` / `getPostSlugs` / `getAdjacentPosts` were removed from `src/content/posts/index.js` with **no** dangling callers — the only remaining import of that module anywhere in `src/` is `src/lib/blog/index.js`.

### 3. The first-run reality — real feed, live database

I ran the actual production path against `https://csalinasdev.substack.com/feed`. It answers **HTTP 200, 1642 bytes, zero `<item>` elements** — the empty-publication case, not a hypothetical.

`node scripts/sync-substack.mjs` against a real MySQL: **exit 0**, `0 seen · 0 created · 0 updated · 0 unchanged · 0 skipped`, `✅ clean` markdown on stdout with nothing else in it, every log line on stderr. No alert fired.

Re-ran the same empty feed against a database that already **had** eight synced rows: still exit 0, and all eight rows survive on `/blog`. Nothing is deleted, nothing is invalidated.

Then the strongest form of "does an empty feed leave the blog exactly as it is today": I built and served `origin/main` in a second clean worktree on another port, captured `/blog` + the three MDX posts with `verify-blog-dom.mjs` from **both** servers (the epic server holding zero synced rows), and diffed:

```
base pages: /blog /blog/goldwater-bank /blog/hashtag /blog/wordleverse
epic pages: /blog /blog/goldwater-bank /blog/hashtag /blog/wordleverse
NO DIFFERENCES
```

Byte-identical structure, text, rail, TOC, footer, cards and article geometry. Then a **negative control** — same base capture, epic server *with* the synced rows seeded — exits **1** with 4 diffs, so the green above is a real measurement rather than a gate that cannot go red.

Failure behaviour, checked by stopping MySQL mid-flight: `/blog` -> 200 (degrades to MDX-only), `/blog/hashtag` -> 200, `/blog/goldwater-bank` -> 200, `/blog/sandwiched` -> 500. That is precisely what `src/lib/blog/index.js` documents: a store outage is a 500, never a `notFound()` that tells a crawler the post is gone.

Deploy plumbing that first run depends on: `.dockerignore` does not exclude `.agent/` (so `esm-resolver.mjs` / `alias-resolver.mjs` ship in the image), and `docker-compose.yml` uses `env_file: .env` for the `app` service, so `SUBSTACK_FEED_URL` reaches `docker compose run --rm app`.

### 4. Coherence of `main`

No stale "the sync must not run on a page render" anywhere. `src/lib/substack/README.md:145` states the rule that survived ("never `await` in a render") and immediately records that the routes *do* call it through `after()`; CLAUDE.md says the same thing in the trigger bullet. The old `src/content/posts/index.js` header claim "the whole blog prerenders" is gone, the `dynamicParams = false` / `generateStaticParams` bullet is rewritten, and the database-seeding recipe exists once (in `seed-substack-fixture.mjs`'s header) with three pointers to it and no drifted copy. The #164 and #167 halves do not contradict each other.

### 5. Build, lint, every verify script

Exit codes captured directly, not through a pipe:

```
DATABASE_URL="mysql://u:p@127.0.0.1:1/none" npm run build   ok   /blog and /blog/[slug] both f (Dynamic)
npm run lint                                                ok   No ESLint warnings or errors
verify-substack-sync.mjs        exit=0
verify-blog-posts.mjs           exit=0
verify-substack-refresh.mjs     exit=0
verify-blog-overflow.mjs        exit=0   (9 pages x 1200/430, no overflow)
verify-blog-dom.mjs --expect    exit=1 on a real difference  (negative control)
```

The throttle behaved in the real server rather than only in the harness: across ~40 requests to `/blog` and eight post pages, the server log contains **exactly one** `[substack] refresh:` pair — one sync, then the 10-minute gate holding.

### Acceptance criteria

- [x] Publishing a post results in a native post after the sync — real feed plus fixtures through the real `prismaStore`
- [x] No Substack iframe/embed — every `<iframe>` becomes a labelled `<a>`; none survives any fixture on a rendered page
- [x] Typography/layout owned by this site — one `dangerouslySetInnerHTML`, on `<Article>` itself, drop cap intact
- [x] Imported HTML sanitized before render — hostile fixtures rendered and inspected: no `javascript:`, no `data:`, no `on*` handler, no relative URL, no protocol-relative URL, no `script`/`svg`/`xmp`/`math`
- [x] Stored locally with Prisma/MySQL — `SubstackPost`, `SubstackSyncStatus`
- [x] Idempotent, updates edited posts — second sync writes zero; `feed-edited` updates in place with the slug frozen
- [x] Existing MDX posts continue to render — DOM identical to `main`
- [x] `/blog` includes both in date order — verified above
- [x] `/blog/[slug]` metadata, hero, reading time, prev/next — verified above
- [x] Available when Substack is unavailable — the sync is post-response; also verified with MySQL down
- [x] Official RSS feed only

### Non-blocking observations

- **`src/app/(pages)/privacy-policy/page.js`** says the post text is one "which my server collects **on a schedule** and stores". There is no schedule — it is visit-triggered behind a 10-minute throttle, plus once per deploy. The material claim (server-side fetch; your browser never contacts Substack for post text) is true and the `substackcdn.com` addition is right; only the word "schedule" is now inaccurate.
- **Three database round-trips per synced post view**: `generateMetadata` -> `getPost` and `Page` -> `getPost` (both `findUnique`, both pulling `contentHtml @db.LongText`), plus `getAdjacentPosts` -> `getPosts` (`findMany`). A React `cache()` wrapper on `getPost` would remove one of the LongText reads. Irrelevant at this site's traffic; noting it because the duplicate is easy to miss.
- An MDX post page still issues the adjacency `findMany`. It degrades correctly (verified 200 with MySQL down), but "MDX posts never touch the database" in CLAUDE.md is a claim about `getPost`, not about the page.
- `verify-blog-overflow.mjs` returned **1** once in ten runs, immediately after another Chrome-spawning gate in the same shell. Nine subsequent runs — including a re-run of the identical loop — were clean and I could not reproduce it or capture the failure, so I am recording it as suspected CDP contention on a busy machine rather than a finding.
- The sanitizer emits `<source ...></source>` (a closing tag on a void element — parsers ignore it) and an empty `<p></p>` where an `<iframe>` had no usable `src`. Both cosmetic.
- CLAUDE.md's `PostBody.jsx` bullet still says the first client render "matches the **prerendered** HTML"; that output is now server-rendered per request. The substance is unchanged.

Nothing here blocks. The gates in this branch are real, they go red when they should, and the one scenario that matters on merge day was executed against the live feed rather than stood in for by a fixture.

PR #173, opened 2026-08-16
refactor(games): plain coloured pieces, initials behind an accessibility toggle

### LINKED ISSUE #172: refactor(games): tile character
For games where the users can't change their name (Connect 404, Race Condition, etc...), don't put the initial on the game piece. Just use the colored game piece. Add an accessibility button in the footer to toggle, that will add a distinct initial into the game piece. But, they should just be colored, empty game pieces.

---

## 🤖 Implementation Plan

_Written by the architect agent on 2026-08-16 10:38 UTC. The implementor builds exactly this._

### Goal

Connect 404 and Race Condition draw **plain coloured pieces with nothing stamped on
them**. A single site-wide accessibility preference — off by default, in
`localStorage`, per browser, invisible to the other player — puts the initial back for
anybody who needs it.

Nothing else changes: the turn banner's chip, the online bar, the endgame panel and the
screen-reader announcements are untouched.

---

### The decision that is being reversed, and the measurement behind it

**Read this before you edit anything, because the code you are changing carries comments
asserting the opposite, and those comments have to be corrected rather than left lying.**

`connect-404/board/Piece.jsx`, `race-condition/board/Marble.jsx` and both `players.js`
headers all state that the initial is load-bearing — "colour never carries a fact on its
own", and for Race Condition (#144) "sixteen marbles sit shoulder to shoulder on a 4x4
and the orbit rearranges all of them at once, so 'which of these is mine' has to survive
a monochrome screen and a moving board." That was a deliberate decision, so it was
tested rather than overridden.

The colours were taken out of the code (`src/app/globals.css`) and run through a
Viénot–Brettel–Mollon dichromat simulation, with WCAG contrast **and** CIEDE2000 (ΔE00 —
the right metric for "can I tell these two adjacent fills apart"; ~2.3 is the just-
noticeable difference, 10+ is "obviously different colours"):

| Game | Vision | ΔE00 between the two players' pieces | piece↔piece contrast | worst piece↔board contrast |
|---|---|---|---|---|
| Connect 404 (`--invalid` #f44747 / `--component` #4fc1ff) | normal | 57.5 | 1.78:1 | 4.34:1 |
| | protanopia | 51.7 | 2.22:1 | 3.65:1 |
| | deuteranopia | 58.6 | 1.29:1 | 5.52:1 |
| | tritanopia | 62.4 | 1.78:1 | 4.34:1 |
| Race Condition (`--component` #4fc1ff / `--parenthesis` #ffd700) | normal | 59.9 | 1.44:1 | 7.74:1 |
| | protanopia | 62.5 | 1.34:1 | 8.09:1 |
| | deuteranopia | 66.9 | 1.63:1 | 7.13:1 |
| | tritanopia | 44.1 | 1.40:1 | 7.74:1 |

Board surface is `--absentBackground` (#28405826) composited over `--background`
(#1f1f1f) = **#202427**; empty sockets are `--background` itself.

**Three conclusions, and they are not all the same answer.**

1. **For the common colour-vision deficiencies, colour alone is comfortably enough.**
   The smallest hue separation anywhere in that table is ΔE00 44, and it is 51+ in every
   red-green case. Both pieces clear 3:1 against the board and the sockets in every
   simulation, so a piece is always distinguishable from an empty square too. Protan,
   deutan and tritan players do not need the initial in either game.

2. **The Race Condition "shoulder to shoulder" premise does not survive reading the
   geometry.** `Marble.jsx`'s `Disc` has `padding: 6%` and the circle is `r=9.2` in a
   `0 0 20 20` viewBox, so a marble is **≈81% of a cell wide** and adjacent marbles are
   separated by ~19% of a cell of dark board plus each marble's own
   `rgba(0,0,0,0.32)` rim. They do not abut. The motion concern from #144 is also
   answered by the hue gap: ΔE00 60 does not become ambiguous because the board turned.

3. **The one case that genuinely fails is luminance-only vision.** Piece-against-piece
   contrast is **1.29–2.22:1** in every row above — below any threshold there is. That
   is achromatopsia, a greyscale display, a monochrome print or screenshot, and
   `forced-colors` mode. It is the "monochrome screen" half of the #144 comment, and it
   is real. It is also the only thing left after this change, because in these two games
   the stamped character is literally the first letter of the colour name (`createPlayers`
   falls back to `colorName` — the players are always called Red/Blue and Cyan/Gold, so
   the initials are R/B and C/G). Removing it removes the board's only non-colour channel
   and takes the games out of conformance with WCAG SC 1.4.1.

**Recommendation: build both halves — plain pieces by default (what was asked for) and
the toggle.** The toggle stops being "if you must": it is what keeps SC 1.4.1 satisfied
once colour is the only per-piece channel, it costs about one small component per game
against a precedent that already exists, and the measurement says it serves a small but
real group rather than a theoretical one.

Two things soften the loss and are worth knowing: the `TurnBanner` still prints the
colour→identity mapping in words at all times ("Red to move", in red, with an R chip),
and the screen-reader `Announcer` in both games narrates by `player.name`, so nothing in
the accessibility tree changes at all — the piece glyphs are already inside
`aria-hidden` subtrees. This is a purely visual change.

**One-sentence overrules, any of which the architect will take without argument:** "ship
it without the toggle"; "put the toggle in the site footer, not the game toolbar"; "use a
ring instead of a filled disc for player two and drop the letter entirely" (that would
fix the greyscale case with no toggle and no character, and was rejected only because
plain coloured pieces were asked for by name).

---

### Scope: which games qualify, and why

The rule is "games where the player cannot change their name", because that is what makes
a stamped character an arbitrary artefact rather than something the player chose.

**In scope — Connect 404 and Race Condition.** Verified, not assumed: neither game has
name entry anywhere. `connect-404/Lobby.jsx` and `race-condition/Lobby.jsx` offer only
"same device" / "play online" / a code field. `race-condition/online/OnlineGame.jsx:68`
builds its cast as `createPlayers([...SEATS])` **with no names at all, deliberately**, and
Connect 404 does the same at `online/OnlineGame.jsx:134`. So `name` is always the colour
name and `initial` is always its first letter.

**Out of scope — Edge Case.** Players type a display name
(`edge-case/online/RoomLobby.jsx`, remembered by `edge-case/online/preferences.js` under
`EDGECASE-NAME`), so its initial is derived from something the player chose. It is
outside the stated rule by construction. It also has up to four players in four colours,
where the argument for a second channel is much stronger.

**Out of scope — Tic Tac Overflow, and this one deserves its reason spelled out.** X and
O are **marks, not initials**. They are not derived from any name — `markForSlot(slot)`
returns them — and there is no `initial` field anywhere in that game (grep it: zero
hits). There is also no colour to be redundant with: the mark *is* the identity, exactly
as in tic-tac-toe on paper. Stripping them leaves two indistinguishable blank cells and
no game. It is not the same change and it is not in this issue.

Wordleverse, Hashtag and Mini Motorways have no per-player pieces.

---

### Approach

Three decisions, each with the loser named.

**1. The preference is a prop threaded from the provider to the piece, not a CSS
variable.** Both boards already thread `players` and `youSlot` down the same path, and
`Game.jsx`'s own comment says "Everything the board needs is computed here and passed
down, which is the whole point". Hiding the glyph with a custom property (the way `--slot`
and `--row` work) would touch two files per game instead of five, but it leaves an
invisible `<text>` node in the DOM and makes "does this board draw initials" ungreppable.
Explicit prop.

**2. The preference lives on the context value as two new keys, not as a reducer action.**
Tic Tac Overflow put `showExpiring` into `state` via a `SET EXPIRY PREVIEW` action, and its
online translator has to intercept that action so it never reaches the wire
(`tic-tac-overflow/online/OnlineGame.jsx:105`). In Connect 404 and Race Condition
`dispatch` **is** the network online — its entire documented contract is "the same actions
in both modes; online turns them into requests". Adding an action that must be remembered
*not* to send is a foot-gun; forget the case and the toggle silently no-ops online, or
worse a display preference reaches room state. Two plain keys on the context value cannot
be sent anywhere. This is the same *shape* as the TTO precedent — `localStorage`, restored
in an after-mount effect because the page is prerendered, per-browser, unable to touch the
other player or the rules — with the one part that is specific to TTO's reducer left out.

**3. One shared preference key across both games, not one per game.** Somebody who needs
initials needs them in every game; making them find the switch twice is a poor
accessibility story. `src/lib/pieceInitials.js` owns the key and both games import it.

**Placement: each game's `Toolbar.jsx`, the strip across the top of the board** — where
Tic Tac Overflow's `ExpiryToggle` already lives. The issue says "footer"; `src/components/Footer.jsx`
is a shared **server** component rendered by `src/app/(pages)/layout.js` on every page in
the site, so a game control there would appear on the blog and could not read game state
without a new app-wide client provider. Toolbar, with the precedent. (Overrulable in one
sentence, as above.)

---

- **There is no test framework in this repo.** `package.json` has `dev`, `build`, `start`,
  `lint`, `sync:substack` and nothing else; there is not a single `.test.js`. Do not add a
  test runner. Verification is lint + build + playing both games in both modes.
- A fresh worktree has no `node_modules`. Run `npm ci` first.
- **The reference component to copy is
  `src/app/(pages)/games/tic-tac-overflow/components/ExpiryToggle.jsx`** — a styled
  `<button type="button">` with `aria-pressed`, a `title`, a dimmed icon at
  `aria-pressed="false"`, and the label "Hint on/off". Match it.
- **The Font Awesome rule from `CLAUDE.md` is live here:** the kit replaces `<i>` with
  `<svg>` in the DOM after mount, so every `<i>` React might unmount must sit inside a
  wrapper element React owns. `ExpiryToggle` and both `Toolbar.jsx` files already do this
  — copy the `<span><i className="…" /></span>` pattern verbatim.
- The storage helper to model is
  `src/app/(pages)/games/edge-case/online/preferences.js`: SSR guard, `try`/`catch` that
  swallows, and a comment saying why a game must not refuse to start because storage was
  unavailable.
- Both `ContextProvider`s note that the page is prerendered and the first client render
  must match the server's markup. That is why the default is `false` and the stored value
  is applied in an effect, never in the initial state.

---

#### New shared module

1. **`src/lib/pieceInitials.js`** (new). Flat in `src/lib/`, beside `qr.js` and
   `rate-limit.js`. Exports:
   - `PIECE_INITIALS_KEY = "CSALINAS-PIECE-INITIALS"`
   - `readPieceInitials()` → `true` / `false` / **`null` when never set** (the caller has
     to tell "never chosen" from "chosen off", exactly as
     `tic-tac-overflow/context/index.jsx:45` does with `saved !== null`). SSR-guarded,
     `try`/`catch`, returns `null` on any failure.
   - `rememberPieceInitials(enabled)` → writes `String(enabled)`, swallows errors.

   Head the file with a comment saying what it is: a per-browser display preference for
   games whose players cannot name themselves, off by default because a piece is a
   coloured piece, on for anybody whose vision is luminance-only — and that it is
   deliberately not room state.

#### Connect 404 — `src/app/(pages)/games/connect-404/`

2. **`board/Piece.jsx`** — add a `showInitials` prop (default `false`); render the
   `<text>` element only when it is true. **Rewrite the JSDoc block above `Piece`** — it
   currently says "The initial is not decoration… the site's rule is that colour never
   carries a fact on its own", which is about to be false. Replace it with the real
   position: red against blue is ΔE00 51+ under every common CVD and clears 3:1 against
   the board, so colour alone carries this board; the initial is behind a preference for
   luminance-only vision. Keep the rim `<circle stroke>` — it is what separates a piece
   from a socket and from its neighbours, and it is doing more work now, not less.
3. **`board/Cell.jsx`** — `Cell` (line 98) takes `showInitials` and passes it to `<Piece>`
   (line 123).
4. **`board/Column.jsx`** — `Column` (line 115) takes `showInitials` and passes it to
   `<Cell>` (line 184). Add it to the JSDoc prop list, which is maintained here.
5. **`board/Board.jsx`** — `Board` (line 41) takes `showInitials` and passes it to both
   `<Column>` (line 221) **and the chute `<Piece>` at line 208** — the piece hovering over
   the column you are about to drop into. Missing that one is the easy bug: it is a
   `<Piece>` rendered directly by `Board`, not through a `Column`.
6. **`Game.jsx`** — read `showInitials` from `useContext(Context)` (line 51) and pass it to
   `<Board>` (line 93). Pass nothing new to `Toolbar` — the toggle reads the context
   itself.
7. **`components/InitialsToggle.jsx`** (new) — modelled line for line on
   `tic-tac-overflow/components/ExpiryToggle.jsx`. Reads `showInitials` and
   `setShowInitials` from the game's `Context`; on click calls `setShowInitials(next)` and
   `rememberPieceInitials(next)`. `aria-pressed={showInitials}`,
   `title="Stamp each piece with its player's initial"`, icon `fa-solid fa-font` inside a
   `<span>`, label `Initials <State>{showInitials ? "on" : "off"}</State>`.
8. **`components/index.js`** — `export * from "./InitialsToggle";`.
9. **`Toolbar.jsx`** — render `<InitialsToggle />` between `<Progress>` and the Restart
   `<Action>`, unconditionally (unlike Restart, it exists in both modes). Give it
   `flex: 0 0 auto` so it survives the narrow-screen layout the file already handles, and
   check it at 480px and 600px where `Name` and the font size change.
10. **`context/index.jsx`** — `ContextProvider` gains `const [showInitials, setShowInitials]
    = useState(false)` and an after-mount `useEffect` that calls `readPieceInitials()` and
    applies it when it is not `null`. Add `showInitials` and `setShowInitials` to the
    `useMemo` store **and to the `createContext` default object**, and document both in the
    provider's comment block: a per-player display preference, never sent, never able to
    change what the other player sees or how the game runs.
11. **`online/OnlineGame.jsx`** — the same `useState(false)` + after-mount effect, and both
    keys added to the `store` `useMemo` (line 132). Add a comment matching
    `tic-tac-overflow/online/OnlineGame.jsx:43-46`: it is deliberately not in the room's
    state, both players set it independently. **Do not add a case to the `dispatch`
    translator** — the setter is not an action and must never reach `send()`.
12. **`players.js`** — correct the header comment. Lines 13–17 currently assert that every
    piece is stamped with its initial. State what is now true: the descriptor still carries
    `initial`, the banner and the endgame panel still use it, and the board draws it only
    when the player has asked for it.

#### Race Condition — `src/app/(pages)/games/race-condition/`

Same nine changes, same reasoning, different component names:

13. **`board/Marble.jsx`** — `showInitials` prop, conditional `<text>`, and **rewrite the
    JSDoc block at lines 84–105**, which is the strongest version of the claim being
    reversed. Say what the measurement found: cyan against gold is ΔE00 44–67 under every
    simulated vision, both marbles clear 7:1 against the board, and the marbles are ≈81%
    of a cell wide so they never actually touch. Keep the rim stroke.
14. **`board/MarbleLayer.jsx`** — `MarbleLayer` (line 44) takes `showInitials`, passes it to
    `<Marble>` (line 64), and documents it in the JSDoc prop list.
15. **`board/Board.jsx`** — `Board` (line 139) takes `showInitials`, passes it to
    `<MarbleLayer>` (line 262).
16. **`Game.jsx`** — read `showInitials` from the context (line 50), pass to `<Board>`
    (line 168).
17. **`components/InitialsToggle.jsx`** (new) + **`components/index.js`** export — same
    component, importing this game's `Context`. It is a near-duplicate of Connect 404's;
    that is consistent with how `Toolbar`, `Panel`, `Button`, `Announcer` and
    `VisuallyHidden` are already duplicated per game in this repo. Do not build a shared
    component directory — only the storage helper is shared.
18. **`Toolbar.jsx`** — render `<InitialsToggle />` alongside Restart, as above.
19. **`context/index.jsx`** — both keys on the store and on the `createContext` default.
    The doc block at lines 13–28 enumerates "the five facts everything below this provider
    is allowed to know" and calls the seam load-bearing — **update that list to seven and
    say what the two new ones are**, or the next person to read it will believe the count.
20. **`online/OnlineGame.jsx`** — `useState` + effect + both keys on the `store` `useMemo`
    (line 156). No `dispatch` case.
21. **`players.js`** — correct the header comment (lines 13–17), which currently carries the
    #144 "sixteen marbles shoulder to shoulder" argument.

#### Documentation

22. **`CLAUDE.md`** — one bullet in "Important Patterns", near the existing Tic Tac
    Overflow `showExpiring` bullet: Connect 404 and Race Condition draw plain coloured
    pieces because neither game lets a player choose a name, so a stamped character was
    always just the first letter of the colour; the initial is behind a per-browser
    preference in `localStorage` under `CSALINAS-PIECE-INITIALS`
    (`src/lib/pieceInitials.js`), off by default, restored in an after-mount effect
    because the pages are prerendered, carried on the context value rather than through
    `dispatch` (which is the wire online) and therefore incapable of reaching room state.
    Note that Edge Case and Tic Tac Overflow are excluded on purpose and why.

---

### Out of scope

- **`TurnBanner.jsx` in either game**, `race-condition/online/OnlineBar.jsx`, and
  `edge-case` anything. Those chips sit next to the player's name in words, they are not
  game pieces, and keeping them is what preserves the colour→identity mapping on screen
  when initials are off.
- `Announcer.jsx` and any `aria-*`. The glyphs are already inside `aria-hidden` subtrees
  and the announcements narrate by `player.name`. Nothing in the accessibility tree
  changes, and nothing there should.
- The palette itself. Do not change `--invalid`, `--component` or `--parenthesis`, and do
  not add a third colour or a pattern fill.
- `player.initial` in `players.js`. It stays on the descriptor in both games; only the
  board stops drawing it.
- `src/components/Footer.jsx`, `src/app/(pages)/layout.js`, and any app-wide preferences
  provider.
- Tic Tac Overflow's `showExpiring` and its `ExpiryToggle`. Read them, copy them, change
  neither.
- No test framework, no new dependency.

---

### Acceptance criteria

- [ ] With no `CSALINAS-PIECE-INITIALS` in `localStorage`, every piece on both boards is a
      plain coloured disc with no character on it — including Connect 404's hovering chute
      piece, Race Condition's sixteen marbles, and the winning-line pieces after a game
      ends.
- [ ] Each game's toolbar has an `Initials off` toggle that turns every piece's initial on
      and back off immediately, in hotseat and online alike.
- [ ] The preference survives a reload, and is shared across the two games: turning it on
      in Connect 404 leaves it on in Race Condition.
- [ ] Playing online in two browsers, one player toggling initials changes **only their own
      screen**. Confirmed by watching the other browser, and by confirming no request is
      made when the toggle is pressed (network tab) and that the room's `revision` does not
      move.
- [ ] `player.initial` is still rendered by `TurnBanner` in both games and by
      `race-condition/online/OnlineBar.jsx`, with or without the preference.
- [ ] No hydration warning in the console on first load of `/games/connect-404` or
      `/games/race-condition`, with the preference set to `true` and to `false`.
- [ ] Edge Case and Tic Tac Overflow are byte-for-byte unchanged apart from nothing —
      neither game is touched.
- [ ] Every comment that previously argued the initial was load-bearing has been corrected
      rather than deleted: `connect-404/board/Piece.jsx`,
      `race-condition/board/Marble.jsx`, and both `players.js` headers now state the
      measured position and the preference.
- [ ] `CLAUDE.md` documents the preference, its key, its default, why it is on the context
      value and not in `dispatch`, and which games are excluded.

### Verification

    npm ci
    npm run lint
    npm run build

Then play both games. There is no test runner in this repo — do not add one.

    npm run dev
    # http://localhost:3000/games/connect-404   -- hotseat, then "Play online" in two
    # http://localhost:3000/games/race-condition   browser profiles against one code

Manual checks, in this order:

1. Fresh profile (clear `localStorage`): pieces are blank. Drop one; the chute piece above
   the column is blank too.
2. Toggle on: initials appear on every piece **and** on the chute piece. Reload: still on.
3. Open the other game: still on. Toggle off there; return to the first game: off.
4. Online, two browsers, one room: toggle in browser A and watch browser B — B's board
   must not change. Check the network tab: pressing the toggle issues no request.
5. Race Condition: press a marble to spin the board with initials on and with them off,
   and confirm the marbles still transition rather than teleport (the `key` is the marble
   id and must not have been disturbed).
6. Narrow the window past 600px and 480px in both games and confirm the toolbar still fits.
7. Keyboard only: the toggle is reachable by Tab, has a visible `:focus-visible` outline,
   and announces its pressed state.

### Risks

- **The chute piece in Connect 404** (`board/Board.jsx:208`) is a `<Piece>` rendered
  directly by `Board`, bypassing `Column`/`Cell`. It is the one that gets missed. You would
  notice as an initial floating above a board of blank pieces.
- **Hydration mismatch** if the stored value is used to seed `useState` instead of applied
  in an effect. Both providers already carry a comment explaining why; the symptom is a
  React hydration warning on a page that looks fine.
- **The toggle silently no-ops in online mode** if one of the two providers is updated and
  the other is not. There are four providers in total (two games × hotseat/online) and all
  four need the same three lines.
- **A display preference leaking into room state** — the thing this design exists to
  prevent. If you find yourself adding a `case` to an `OnlineGame` `dispatch` switch, stop:
  the setter is not an action.
- **The toolbar overflowing on a phone.** Both `Toolbar.jsx` files already drop the game
  name below 480px and shrink the font below 600px; the new button is a third flex child
  and needs `flex: 0 0 auto` plus a check at those widths.
- **Comment drift.** Four comment blocks currently argue the opposite of this change. Left
  in place they will send the next agent — or the next reviewer — to the wrong conclusion,
  which is exactly how the #144 reasoning survived long enough to need measuring.

csalinas-dev:
Closes #172

Built to Christopher's 2026-08-16 decision on the issue: keep the theme, remove the
initial by default, add one accessibility toggle that puts the same initial back. The
toggle's only job is to show or hide the glyph that already exists — nothing is
restyled, resized, recoloured, ringed or given a second channel.

## What changed

- **`src/lib/pieceInitials.js`** (new, 42 lines) — the one preference key,
  `CSALINAS-PIECE-INITIALS`. `readPieceInitials()` returns `true`/`false`/`null`, because
  the caller has to tell "never chosen" from "chosen off". SSR-guarded, errors swallowed,
  modelled on `edge-case/online/preferences.js`.
- **`connect-404/board/Piece.jsx`, `race-condition/board/Marble.jsx`** — a `showInitials`
  prop gates the existing `<text>`. The element itself is untouched: still `font-size="9"`,
  `x="10" y="10.4"`, `text-anchor="middle"`, `fill: var(--background)`, `font-weight: 700`.
  Both rim strokes stay.
- **The prop is threaded down** `Board → Column → Cell → Piece` and `Board → MarbleLayer →
  Marble`, from `Game.jsx` reading the context. Connect 404's **chute piece** — the one
  `Board` renders directly, bypassing `Column`/`Cell` — is included; it is the one the plan
  flagged as easy to miss.
- **Two `components/InitialsToggle.jsx`**, one per game, copied from
  `tic-tac-overflow/components/ExpiryToggle.jsx`: `aria-pressed`, a `title`, the icon dimmed
  when off, `<i>` inside a `<span>` React owns. Rendered unconditionally in each game's
  `Toolbar` between the progress count and Restart (Restart is absent online; this is not).
- **All four providers** — `context/index.jsx` and `online/OnlineGame.jsx` in both games —
  gain the same `useState(false)` plus an after-mount effect, and put `showInitials` /
  `setShowInitials` on the context value. **No `dispatch` case was added to either online
  translator**; the setter is not an action.
- **Four comment blocks corrected, not deleted** — `Piece.jsx`, `Marble.jsx` and both
  `players.js` headers. They now carry the measurement, including the finding that retired
  the #144 rationale: `Disc` has `padding: 6%` and the circle is `r=9.2` in a `0 0 20 20`
  viewBox, so a marble is ~81% of a square wide and adjacent marbles are separated by ~19%
  of a square of board plus each marble's own rim. **They do not touch.** (I also corrected
  a fifth block, the `cast` comment in `race-condition/online/OnlineGame.jsx`, which
  repeated the same "sixteen of them touch" claim. Its actual point — that nameless seats
  must not both initial as "P" — is kept, and now matters *more*, since a player who turns
  the preference on has to see a C and a G.)
- **`CLAUDE.md`** — one bullet beside the existing `showExpiring` one.

## How it was verified

Automated with `.agent/scripts/shot.mjs`, which exits 3 when a page script reports
failure. Probes were scratch files and are not in the diff.

    npm ci && npx prisma generate
    npm run lint     -> No ESLint warnings or errors
    npm run build    -> compiled successfully, both game routes prerendered

Connect 404, hotseat, four pieces played, 1200px:

    {"ok": true, "offLetters": 0, "offPressed": "false",
     "onLetters": ["R","B","R","B"], "onPressed": "true", "stored": "true",
     "backOffLetters": 0, "storedOff": "false",
     "geom": {"fontSize":"9","x":"10","y":"10.4","textAnchor":"middle",
              "fill":"rgb(31, 31, 31)","fontWeight":"700"},
     "disc": {"r":"8.6","stroke":"rgba(0, 0, 0, 0.32)"}}

The `geom`/`disc` block is the "no other alterations" check: those are `main`'s values.
Race Condition returned the same, with `r: 9.2` and letters `C`/`G`, and the marble count
was identical before and after toggling — the `key` is still the marble id, so the orbit
still transitions rather than teleports.

Chute piece and restore-after-mount:

    {"ok": true, "restored": "true", "chuteLetterOn": 1,
     "chuteStillADisc": true, "chuteLetterOff": 0,
     "focusable": true, "consoleErrors": []}

Toolbar layout, both games at 430 / 480 / 600 / 1200px: `ok=true`, zero children outside
the strip's content box, `scrollWidth == clientWidth` at every width, toggle 32px tall.

Screenshots looked at directly: plain discs at 1200px and 430px in both games; initials
on in both; the `TurnBanner` chip still shows `R`/`C` in both states, as required.

## Notes for the reviewer

- **I could not open a real room, and this is the gap in the verification.** The
  worktree's provisioned `.env` carries a Postgres placeholder for `DATABASE_URL`, so
  `POST /api/rooms` returns 500 and no online board can be reached here. I did not go
  looking for other credentials. So the acceptance criterion "two browsers, one toggles,
  the other's board does not move / no request / revision does not move" is **not**
  confirmed on screen by me — please confirm it, or run it where a MySQL is configured.
  What I did instead is a static check with a mutation test: it asserts all four store
  literals carry both keys, that neither toggle mentions `dispatch`/`send`/`fetch`/
  `useRoom`, and that neither online translator gained a case. Its first version passed
  against a provider I had deliberately broken — a whole-file `includes()` matched the
  `useState` declaration — so I tightened it to read inside the `useMemo` store literal
  and re-confirmed it fails on the mutation before restoring the file. Worth knowing that
  the weak version existed.
- **One deviation from `ExpiryToggle`, and it is deliberate:** `min-height: 2rem`, the same
  as the Restart button beside it. Copied line for line the toggle is a 16px-tall tap
  target, under WCAG 2.2 SC 2.5.8, which is a poor thing for the one control on the screen
  that exists for accessibility. It costs no layout — Restart already sets the toolbar's
  height in hotseat. It does make the *online* toolbar (no Restart) 32px tall where it was
  shorter on `main`. Say the word and I will drop it.
- **The plan's hydration risk does not exist on these pages, but the effect is still
  right.** `Connect404.jsx` / the Race Condition equivalent prerender a *Lobby*; the board
  and its provider only mount after "same device" is pressed, well after hydration. So
  seeding `useState` from storage could not have produced a mismatch here. The
  after-mount effect is kept anyway — it matches the TTO precedent and it stays correct if
  the mode ever becomes a route.
- **`player.initial` is untouched on the descriptor**, and `TurnBanner` in both games plus
  `race-condition/online/OnlineBar.jsx` still render it in both states. That is what keeps
  the colour→identity mapping on screen when the initials are off.
- Edge Case and Tic Tac Overflow are not touched at all.


csalinas-dev:
## 🤖 Review: APPROVED

Checked security, correctness, efficiency and fit. I checked the branch out, built it,
and ran it against a second worktree of `main` on a neighbouring port so every claim
below is a comparison, not a reading of the diff.

### 1. The toggle only shows and hides the glyph that already existed

Confirmed independently of the implementor's DOM check, and more strongly: **a pixel
diff of the board region between `main` and this branch with initials on.**

| Board region | Pixels compared | Pixels differing | Max channel delta |
|---|---|---|---|
| Connect 404, `main` vs PR **initials on** | 197,760 | **0** | 0 |
| Race Condition, `main` vs PR **initials on** | 173,056 | **0** | 0 |
| Connect 404, `main` vs PR **initials off** (control) | 197,760 | 1,665 | 224 |
| Race Condition, `main` vs PR **initials off** (control) | 173,056 | 6,021 | 224 |

Same deterministic position on both sides (neither game contains any randomness), the
same probe script driving both, and a crop anchor asserting the region is real content
rather than backdrop. The controls are what make the zeroes mean something — the same
tool finds the letters when they are absent.

Alongside that, a structural dump of every piece: stripped SVG markup, element
attributes, computed styles (`fill`, `stroke`, `stroke-width`, `font-family`,
`font-weight`, `font-size`, `text-anchor`, `dominant-baseline`) and bounding rects to
0.01px. Between `main` and initials-on: **identical in every field**. Between initials-on
and initials-off, the *only* fields that differ are `html` and `text` — the disc, the rim
(`r=8.6` / `r=9.2`, `rgba(0, 0, 0, 0.32)`, `stroke-width 0.7`), the `.disc` rect, the svg
rect and the board origin are byte-for-byte the same in both states. The board origin is
also unchanged from `main`, so the extra toolbar button shifts nothing.

**One thing I chased down and want on the record, because it looks like a finding and is
not.** Toggling mid-session, Race Condition's letter ink box measures 24.03x46 where
`main`'s measures 25x45 — reproducibly. It is not this PR. On **`main`**, cloning the
existing `<text>` and re-inserting it produces exactly the same 24.03x46. It is a
headless-Chrome artifact of a text node created late, present on `main` too. On the real
returning-visitor path (preference already stored, letters present from the marble's
first render) the PR measures 25x45 at the same y — identical to `main`, which is the
pixel-diff row above.

Winning-line pieces, which get their own ring and `pop` animation, obey the preference in
both directions (4/4 blank off, 4/4 lettered on, "Red wins" banner intact). Connect 404's
chute piece — the `<Piece>` `Board` renders directly — is blank when off and carries `R`
when on.

### 2. Default is off, per browser, cannot reach the other player

Fresh profile, no stored key: all six pieces `null` in both games. Confirmed on screen.

For "cannot reach room state" I wrote my own static check rather than trusting the
implementor's, and mutation-tested it: for each of the four providers it brace-matches the
object literal passed to `const store = useMemo(` — the value actually handed to
`Context.Provider` — and then re-runs the same assertion against a copy with the key
deleted, failing loudly if the assertion still passes. All four mutations flip the check.
It also confirms neither toggle mentions `dispatch`, `send(`, `fetch(` or `useRoom`,
neither online translator gained an `INITIAL` case, no provider seeds `useState` from
storage, and every provider has the after-mount `readPieceInitials()` effect. A sweep
anchor fails the run if fewer than four mutation tests actually execute.

Worth knowing: my first version of that check had the same class of false green the
implementor reported — it anchored on `useMemo(` and matched `const cast = useMemo(...)`
in Race Condition's `OnlineGame`, reporting the keys missing. Anchoring on
`const store = useMemo(` fixed it. Both online paths render the same shared `<Game>` under
the same `Context.Provider`, so the toolbar toggle and the board's `showInitials` come off
the literal that was mutation-tested.

### 3. Survives a reload, and one key serves both games

Driven end to end in one browser profile, using same-origin iframes so each step is a
genuinely fresh document load re-running the provider's effect against `localStorage`:

1. Connect 404, nothing stored -> pieces blank
2. toggle on -> `R B R`, `aria-pressed="true"`, stored `"true"`
3. **fresh load of Race Condition** -> comes up on, `C G C`, toggle reads `"true"`
4. toggle off there -> marbles blank, stored `"false"`
5. **fresh load of Connect 404** -> comes up off, toggle reads `"false"`

Both directions, one key, `CSALINAS-PIECE-INITIALS`.

### 4. Scope

23 files, all of them the two games' board/toolbar/context files, the two toggles, the
shared helper and one `CLAUDE.md` bullet. `git diff --name-only origin/main...HEAD` over
`edge-case`, `tic-tac-overflow`, `src/lib/realtime`, `prisma`, `package.json` and
`package-lock.json` returns **zero files**. No new dependency. The four comment blocks
that argued the initial was load-bearing are corrected rather than deleted, plus the
`cast` comment in `race-condition/online/OnlineGame.jsx` that repeated the same claim.

### Also checked

- `npm run lint` — no warnings or errors. `npm run build` — compiled, both game routes
  still prerendered (`○`).
- Toolbar at **430 / 480 / 600 / 1200px**, both games, both preference states: no child
  spills the strip, `scrollWidth == clientWidth`, toggle focusable, and **zero console
  errors or warnings** — no hydration warning with the preference set to `true` or to
  `false`.
- `player.initial` still rendered by both `TurnBanner`s (the `R` and `C` chips are visible
  with initials off) and untouched on the descriptor.
- Nothing security-relevant: no network, no new dependency, no user-controlled string
  reaches the DOM. `readPieceInitials` is SSR-guarded and coerces through `=== "true"`.

### The online gap, closed as far as it can be here

I reproduced it rather than taking it on trust: `POST /api/rooms` returns 500 and the
server log says `Error validating datasource db: the URL must start with the protocol
mysql://`. The provisioned `DATABASE_URL` is the wrong engine, so no room can be opened in
this worktree. Not blocking — both games work in production and this change gates a
`<text>` element behind a context key that four mutation-tested assertions say never
reaches `dispatch`.

### Non-blocking

- `connect-404/board/Cell.jsx:99` is now an 84-character single-line parameter list that
  default Prettier would wrap. The repo has no Prettier config, no Prettier dependency and
  no format script, and `Marble.jsx` already failed the same check on `main`, so nothing
  enforces it — noting it only so it is a choice rather than an accident.
- The preference does not sync across tabs until reload (no `storage` listener). Same as
  the `TICTACOVERFLOW-PREVIEW` precedent; mentioning it, not asking for it.

Right at the size it was asked for. Ship it.

(Posted as a comment rather than an approving review — GitHub will not let the PR's own
author account approve it.)

Issue #154, opened 2026-08-16
feat(blog): scheduled sync trigger, auth, and operational reporting


csalinas-dev:
Part 3 of epic #151. **Base branch: `epic/substack-rss`.** Depends on parts 1 and 2.

Wire the sync function from part 1 to something that actually runs, per **section 5** of #151 ("callable from a deployment/cron mechanism"). Read #151 for the requirement; this issue only adds the constraints that come from this repo.

- **If the trigger is an HTTP route, it is an unauthenticated write endpoint until you make it otherwise.** This site already has the pattern for a gated route: `src/app/api/github/card` uses a shared secret plus `src/lib/rate-limit.js`, and the reasoning is written up in the repo's CLAUDE.md. Follow it, or justify something better. An endpoint that lets an anonymous caller make the server fetch a URL and write rows is worth a moment's thought.
- Failures must be visible. A sync that silently stops working leaves the blog looking fine and quietly frozen — decide how a failure surfaces and say so in the PR.
- If this needs anything in `.github/workflows/deploy.yml`, remember that step ordering matters there: the schema push runs after the image build.

### Done when
- The sync runs on a schedule or deploy without a human invoking it.
- Any HTTP surface is authenticated and rate-limited, and the reviewer can demonstrate an unauthenticated call is refused.
- A failed sync is observable rather than silent.
- The mechanism, its secret, and how to run it by hand are documented in the repo's CLAUDE.md.

---

## 🤖 Implementation Plan

_Written by the architect agent on 2026-08-16 04:36 UTC. The implementor builds exactly this._

### Goal

`syncSubstackPosts` runs on its own — every 30 minutes and after every successful
deploy — without anybody invoking it, and a run that fails is impossible to miss:
it turns the workflow run red and opens a GitHub issue that closes itself when the
next run succeeds. **No new HTTP endpoint exists**, so there is nothing on the
public internet to authenticate, rate-limit, or abuse.

### Approach

**The trigger is a scheduled GitHub Actions workflow on the same self-hosted
runner that already deploys the site**, invoking the sync in a throwaway container:
`docker compose run --rm app node scripts/sync-substack.mjs`. That is exactly the
shape `.github/workflows/deploy.yml` already uses for `prisma db push`.

*Why not an HTTP route.* The issue points at `src/app/api/github/card` as the
pattern for gating a route, and that pattern is right — for that route. The card
route **has to** be public: GitHub's Camo proxy fetches it from the internet, so a
[...]

Issue #43, opened 2024-02-23
feat(hashtag): calculate letter statuses


csalinas-dev:
Calculate yellow and green letters.
```
Grey - not in word
Yellow - in word, not in right spot
Green - in word and in correct spot
```

Issue #42, opened 2024-02-23
feat(hashtag): draggable


csalinas-dev:
Use the `draggable` attribute. [Draggable](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/draggable)

PR #175, opened 2026-08-17
fix(deploy): stop the deploy referencing an action this repo forbids

### LINKED ISSUE #174: fix(deploy): every deploy since the Substack epic has ended in startup_failure — nothing has shipped since 06:35
**Production has not deployed since 2026-08-16T06:35Z.** Three consecutive `Deploy csalinas.dev` runs have ended in `startup_failure`, which means GitHub could not parse or start the workflow at all — no step ran, so there is no step log to read.

| Run | Time | Result | Merge that triggered it |
|---|---|---|---|
| 31941075724 | 10:12:43Z | startup_failure | #169, the Substack epic |
| 31942082108 | 10:35:50Z | startup_failure | #171, privacy wording |
| 31977743519 | 22:56:51Z | startup_failure | #173, piece initials |

The last success was 06:35Z (#160). **The first failure is 2 seconds after the Substack epic merged**, and the epic is what last edited `.github/workflows/deploy.yml` — #154 added a "Sync blog from Substack" step and two `github-script` steps that open and close an alert issue.

So `deploy.yml` almost certainly does not parse, or contains an expression GitHub rejects at load time. The two later failures are the same broken file being re-run; they are not three separate bugs.

### What this means right now

Everything merged today is built and reviewed but **not live**: the whole Substack feature, the privacy-policy correction, and the piece-initials change. The site is serving the 06:35 build.

### How it got through

Worth stating plainly so the fix does not stop at the symptom. The reviewer of #154 dry-ran both `github-script` bodies against a stubbed Octokit and confirmed all six branches behaved — good work that could not have caught this, because it tested the JavaScript *inside* the steps and never validated the workflow file itself. Nothing in this repo parses `deploy.yml`, and a `startup_failure` produces no step output to notice.

### Scope

- Make the workflow parse and run again. **Do not change what it does** — the deploy behaviour, step ordering and the sync step were reviewed and are not suspect.
- If the alert steps are what break it, fix the syntax rather than deleting them; they are the failure channel #154 exists to provide.
- Add something that would have caught this. A parse/lint of the workflow file is the obvious candidate, but it must be something that actually runs — this repo has no CI, and a check nobody invokes is not a check.

### Done when
- A push to `main` produces a run that reaches its first step.
- A deploy completes and the site serves the current `main`.
- Something in the repo would fail on an unparseable workflow file, and has been shown to fail on one.

csalinas-dev:
Closes #174

## The diagnosis first, because it is not what the issue assumed

**`deploy.yml` parses fine. There is no syntax error.** I checked before changing anything:

| Validator | Verdict on the broken file at `8c5a57e` |
|---|---|
| `actionlint` v1.7.7 | clean (only the expected `label "sfo3-02" is unknown` false positive) |
| `@actions/workflow-parser` v0.3.61 — **GitHub's own parser**, the one behind the web editor and the VS Code extension | `OK: no errors` |
| byte scan (BOM / CRLF / NUL / tabs / non-ASCII) | clean; the only non-ASCII are intended em-dashes and one 🤖 |

I confirmed that harness can go red before trusting it green — it correctly rejects a mis-indented sequence item and `issue:` for `issues:`.

**The actual defect is a repository policy, not a file defect:**

```
$ gh api repos/csalinas-dev/csalinas-dev-site/actions/permissions
{"enabled":true,"allowed_actions":"local_only","sha_pinning_required":false}
```

`local_only` permits only actions defined **inside this repository**. #154 added two `uses: actions/github-script@v7` steps — the *first* `uses:` this repo has ever had (`git log -p --all -- .github/workflows/` confirms every earlier step was `run:`). GitHub refuses such a workflow at **load** time, so no job is created, no step runs, and there is no annotation or step log. That is precisely the `startup_failure` signature, and it is why the first failure is two seconds behind that merge.

### Proved, not inferred

I pushed a throwaway branch with two minimal workflows **in one commit**, differing in exactly one respect. Both were accepted by actionlint *and* GitHub's own parser:

| Probe | Only distinguishing content | Result |
|---|---|---|
| `probe-174-uses` | `uses: actions/github-script@v7` | **`startup_failure`** |
| `probe-174-run` | `run: echo hello` | **`success`** |

A third probe on the real `[self-hosted, sfo3-02]` box inventoried it: `curl`, `jq`, `docker`, `python3` present; **`gh` and host-level `node` absent**. So the fix below does not guess at what it can call. The probe branch is deleted; the three runs remain in Actions history and can be ignored.

## What changed

- **`.github/workflows/deploy.yml`** — the two alert steps become `run:` steps using `curl` + `jq` against the REST API instead of `actions/github-script`. **Behaviour is unchanged**: same `ops`-label + exact-title lookup, same `Still failing — <run>` comment when one is open, same new issue with labels `["bug","ops"]` and the same body, same `Sync recovered — <run>` comment and close. `continue-on-error` stays off both, the `steps.sync.outcome` keying is untouched, step order is untouched, and the `permissions: issues: write` block stays (the repo default is `read`). Nothing else in the file moved.
- **`.agent/scripts/verify-workflows.mjs`** — the gate that was missing. Two rules: every workflow parses as YAML, and **every `uses:` is `./`-local**. It walks the *parsed* tree rather than grepping, so a `uses:` written inside a `run: |` block is correctly ignored.
- **`.agent/fixtures/workflows/`** — one known-bad file per rule (`external-action.yml` is the real #154 defect reduced) plus a known-good one that deliberately contains the string `uses: actions/github-script@v7` inside a `run:` block.
- **`.github/workflows/verify-workflows.yml`** — runs the gate on every PR touching a workflow. **It uses no actions either**: `actions/checkout` is itself forbidden here, so it fetches the tree with `curl` on the tarball endpoint. A gate that `startup_failure`d for the fault it exists to catch would be worse than no gate.
- **`package.json`** — `js-yaml` promoted to an explicit devDependency (it was already in the tree transitively). One line; the lockfile diff is one line.
- **`CLAUDE.md`** — one bullet stating the rule, why the policy exists, what the runner has, and that a linter will not catch this.

## How it was verified

The gate goes red on the **real** broken file, not just a fixture. I committed first, then restored `main`'s `deploy.yml` over it:

    $ git checkout 8c5a57e -- .github/workflows/deploy.yml
    anchor: github-script references present = 2
    $ node .agent/scripts/verify-workflows.mjs
    self-test ok  (3 fixtures: 2 must fail, 1 must pass)
    FAIL  .github/workflows/deploy.yml: jobs.deploy.steps[5].uses references "actions/github-script@v7", ...
    FAIL  .github/workflows/deploy.yml: jobs.deploy.steps[6].uses references "actions/github-script@v7", ...
    2 violation(s) across 2 workflow file(s).
    EXIT=1

    $ git checkout HEAD -- .github/workflows/deploy.yml   # explicit HEAD, not `--`
    $ git diff HEAD --quiet && git diff --cached --quiet  # both silent
    $ node .agent/scripts/verify-workflows.mjs
    self-test ok  (3 fixtures: 2 must fail, 1 must pass)
    ok    .github/workflows/deploy.yml
    ok    .github/workflows/verify-workflows.yml
    EXIT=0

The self-test is permanent, not a one-off: if a must-fail fixture ever passes, the script exits non-zero **without reporting on the real workflows at all**.

Other checks, all green:

    actionlint + @actions/workflow-parser on both workflow files -> clean / OK: no errors
    bash -n on all 8 run: blocks in deploy.yml (anchored at >=5)  -> all syntax ok
    jq 1.7.1 on the selector: 3-item list -> 42; no match -> empty; [] -> empty (exit 0)
    jq --arg round trip of a report with newlines, quotes, backslashes, pipes, em-dash
      -> valid JSON, byte-identical back out (69 bytes)
    empty-report fallback sentence fires on "" and on whitespace-only
    npm run lint -> No ESLint warnings or errors

## Notes for the reviewer

**What only a real run can confirm.** I cannot prove the deploy succeeds — only GitHub can accept the file, and only the sfo3-02 box can run it. Specifically **unverified until this merges**:

- that GitHub actually loads the new `deploy.yml` (my evidence is three validators plus the probe showing a `run:`-only workflow starts — strong, but not the same as this file starting);
- the two alert steps' *live* REST behaviour. I tested the jq filters and JSON escaping offline against canned payloads, but no call has been made to the real API. They fire only when the Substack sync fails or recovers, so a green deploy will exercise **neither** — the first real test of the failure path is the first real sync failure;
- `verify-workflows.yml` itself has never run. The tarball-instead-of-checkout step is the novel part.

**Things I'd flag on myself:**

- The alert steps now fail the run if `curl --fail` gets a non-2xx (`set -euo pipefail`). `github-script` also threw on API errors, so this matches — and #154's comment says a broken alarm is *worth* a red run. But it is a new way for a deploy to go red **after the site is already up**, and it is deliberate.
- The REST list endpoint returns pull requests alongside issues. `issues.listForRepo` did too, so this is unchanged, and the exact-title match disambiguates.
- I did **not** touch `allowed_actions`. Flipping it to `selected` with `actions/*` would be a one-call fix with a zero-line diff and would keep the reviewed steps verbatim — but it lets third-party action code execute on Christopher's own self-hosted server, which is very likely why `local_only` was set. That is his call, not mine; if he'd rather loosen the policy, this PR's `deploy.yml` change can be reverted and the gate kept.
- The gate hardcodes the `local_only` expectation rather than querying the API, to stay offline like every other `.agent/scripts` check. If the policy is ever loosened, that rule needs relaxing with it — the script says so in its header.
- `verify-workflows.yml` installs only `js-yaml` rather than running `npm ci`, to keep the job to seconds. That means the gate's one dependency is pinned in two places (`package.json` and that workflow).

https://claude.ai/code/session_01Gk1KrpkjCKpJ4uCKfocG9b


csalinas-dev:
### Update — two claims in the body above are now stale, in the PR's favour

**1. `verify-workflows.yml` is no longer unverified.** It ran on this very PR and passed, twice:

- run [31982507549](https://github.com/csalinas-dev/csalinas-dev-site/actions/runs/31982507549) and [31982570682](https://github.com/csalinas-dev/csalinas-dev-site/actions/runs/31982570682) — both `success`

```
self-test ok  (3 fixtures: 2 must fail, 1 must pass)
ok    .github/workflows/deploy.yml
ok    .github/workflows/verify-workflows.yml
2 workflow file(s) ok, 0 uses: reference(s) checked.
```

So the tarball-instead-of-`actions/checkout` step works, and — worth stating plainly — **this is a second, independent confirmation of the diagnosis**: a workflow using no external actions loads and runs in this repo today, on the same commit where `deploy.yml`'s `uses:` version would not.

**2. I was wrong about the install step, and CI caught it.** I claimed `npm install --no-save js-yaml@4.1.1` would keep the job "to seconds" by installing one package. The log says `added 599 packages in 23s` — npm still reads the repo's `package.json` and installs the whole tree, so naming the package bought nothing. Fixed in `c58b3fa`: it is now `npm ci --no-audit --no-fund`, which costs the same 22s and is at least reproducible against the lockfile. The misleading comment is corrected rather than left in.

### What is still unverified

Unchanged, and still the important caveat: **I cannot prove the deploy succeeds.** Only a push to `main` will show GitHub accepting `deploy.yml`, and only the sfo3-02 box will run it. And the two alert steps' live REST behaviour stays untested until a Substack sync actually fails or recovers — a green deploy exercises neither path.

csalinas-dev:
## 🤖 Review: APPROVED

Checked: security, correctness, efficiency, fit against the plan. I re-derived the
diagnosis from scratch rather than taking it on trust, and I ran both rewritten
alert steps end to end against a stubbed API, because a green deploy exercises
neither of them.

### The diagnosis is right — verified three independent ways

1. **Policy.** `gh api repos/csalinas-dev/csalinas-dev-site/actions/permissions`
   → `{"enabled":true,"allowed_actions":"local_only","sha_pinning_required":false}`.
   `.../permissions/selected-actions` answers **409 Conflict: "Only local actions
   and workflows are allowed on this repository"**, which is the policy confirming
   itself from a second endpoint. The repo owner is a User account, so there is no
   org policy above it doing something else.
2. **The probe is real and is in the run history.** `probe-174-uses` →
   `startup_failure`, `probe-174-run` → `success`, `probe-174-tools` → `success`,
   all on `probe/issue-174-local-only`, adjacent run ids from one push. The
   controlled variable really is the `uses:`.
3. **Not a parse error.** I installed `@actions/workflow-parser@0.3.61` myself and
   ran it over `main`'s `deploy.yml` at `8c5a57e`: **`OK`, no errors**. Same
   harness reports errors on `unparseable.yml` (`All sequence items must start at
   the same column at line 10, column 8`), so it can go red. The issue's premise —
   "`deploy.yml` almost certainly does not parse" — is wrong, and the PR is right
   to overturn it. Both new workflow files also parse `OK` under that parser.

### The rewritten alert steps — behaviour is preserved

I extracted both `run:` blocks out of the parsed YAML (so I ran exactly what the
runner will run), put a recording stub `curl` and a real `jq 1.7.1` on `PATH`, and
drove every branch:

| Scenario | Requests made | Verdict |
|---|---|---|
| open, `[]` from the list | GET, then `POST /issues` with `{"title":"Substack sync is failing","labels":["bug","ops"],"body":…}` | matches `issues.create` |
| open, list has `"Substack sync is failing (old)"` and an unrelated issue | GET, then `POST /issues` | **exact** title match preserved — a superset title does not match, same as `=== TITLE` |
| open, list has #9, #42, #43 (two exact matches) | GET, then `POST /issues/42/comments`, no create | `.[0]` picks the same element `.find()` would; list order and `per_page=100` unchanged |
| open, report file absent | create, body ends `_The sync produced no summary — …_` | matches the `try/catch` + falsy fallback |
| open, report file whitespace-only | same fallback | matches `.trim()` falsiness |
| close, no match | GET only, exit 0 | matches `if (!existing) return` |
| close, match | `POST /issues/42/comments` **then** `PATCH /issues/42 {"state":"closed"}` | comment-before-close ordering preserved |

**Awkward payload.** I wrote a report containing `"` `\` tab, newlines, `$VAR`,
backticks, `%s`, `${x}`, an em dash, `é中文` and an emoji, then parsed the POST
body back with `JSON.parse` and compared to the file byte for byte: **exact
round-trip**. No shell expansion, no `printf` format interpretation (all user data
goes through `%s` arguments, never the format string), correct JSON escaping from
`jq --arg`. The empty-array case is safe too: `[] | .[0].number // empty` yields
`null.number` → `null` → `empty`, not an error.

**Error paths.** With the stub returning non-2xx: `--fail` + `set -euo pipefail`
propagates on the GET (exit 22), on the create/comment POST (22) and on the close
PATCH (22); a non-JSON list body fails in `jq` (exit 5). So a broken alarm is a red
run, which is what #154's comment asked for and what `github-script` did by
throwing. Nothing is swallowed and no half-state is possible beyond
"commented but not closed", which the original had too.

`steps.sync.outcome` keying, step order, absence of `continue-on-error` on both
alert steps, and `permissions: { contents: read, issues: write }` are all
untouched. Only `curl` and `jq` are invoked — no `gh`, no host `node` — matching
the probed inventory.

### The gate

- **Shown red on the real file, not a synthetic one.** I restored `main`'s
  `deploy.yml` at `8c5a57e` over the branch's and ran it: two `FAIL`s naming
  `jobs.deploy.steps[5].uses` and `steps[6].uses`, `EXIT=1`. Restored with
  `git checkout HEAD --` and `git diff HEAD --quiet` confirmed clean.
- **No false positive on `uses:` inside `run:`.** I added a scratch workflow whose
  `run: |` block contains both `echo "uses: actions/checkout@v4"` and a
  `# uses: actions/github-script@v7` comment — **passes**. It walks the parsed
  tree, so this is structural, not a grep that happens to work.
- **No false negative on the shapes that matter.** Job-level reusable-workflow
  `uses: octo/other/.github/workflows/x.yml@main` → rejected;
  `uses: docker://alpine:3.19` → rejected; two `./`-local composite refs → accepted
  and counted.
- **The self-test is the anchor** and it is load-bearing: if `collectUses` ever
  stopped finding anything, `external-action.yml` would pass and the script exits
  without reporting on the real workflows at all.
- `verify-workflows.yml` **has** run — twice, green, 27s. I read the job log: the
  tarball step really did deliver `.agent/fixtures/`, `npm ci` installed 599
  packages, and the gate printed `self-test ok` plus both real workflows. It uses
  no actions, so it is not subject to the fault it exists to catch.
- `js-yaml@4.1.1` is a single lockfile entry, `dev: true`, already present
  transitively — the lock diff is one line and no new package is downloaded.
  `.agent/scripts/verify-*.mjs` is the established idiom here (11 siblings).
  `npm run lint` clean.

### On not flipping `allowed_actions`

**Agreed, and strongly.** `selected` with `actions/*` would mean tag-mutable
third-party code executing on the self-hosted `sfo3-02` box that also holds the
production containers and the database — a supply-chain compromise of any allowed
action becomes root-adjacent on Christopher's server. The policy is also the only
thing that has ever caught this class of mistake here. Keeping it and paying ~40
lines of shell is the right trade.

### Acceptance criteria

- [x] *Make the workflow parse and run again* — the only load-time defect (`uses:`
      pointing outside the repo) is gone; both files parse under GitHub's own
      parser and the gate reports zero non-local `uses:`.
- [x] *Do not change what it does* — deploy steps, ordering, `id: sync`,
      `continue-on-error`, `outcome` keying and `permissions` are byte-identical
      apart from the two rewritten step bodies, whose behaviour I exercised
      branch by branch above.
- [x] *Fix the alert steps rather than delete them* — both survive intact.
- [x] *Add something that would have caught this, and show it failing on one* —
      `verify-workflows.mjs` fails on `main`'s actual broken `deploy.yml`
      (reproduced independently above) and runs in CI on every workflow-touching PR.
- [ ] *A push to `main` reaches its first step / a deploy completes* — only
      merging can show this, and the PR says so plainly instead of claiming
      otherwise. That is the correct thing to have done.

### Non-blocking observations

1. **The gate does not scan `.github/actions/**/action.yml`.** CLAUDE.md and the
   failure message both point people at "add a `./`-local composite action" as the
   remedy — and a composite action that itself carries
   `uses: actions/checkout@v4` would sail past this gate and reproduce the outage.
   No such directory exists today, so this is a follow-up, not a blocker; worth
   widening the scan when the first local action is written.
2. **No branch protection on `main`** (`/branches/main/protection` → 404, no
   rulesets), so `Verify workflows` is advisory rather than required. The issue
   asked for a check that fails, and it does — but if you want it to actually
   *stop* a merge, that is a repo setting only you can flip.
3. **Two stale lines in the PR body**, not in the code: it says
   `verify-workflows.yml` "has never run" (it has, green, twice) and that the job
   "installs only `js-yaml` rather than running `npm ci`" — commit `f0cfb91`
   changed that to `npm ci`, so the "pinned in two places" caveat no longer
   applies. Nothing to fix in the tree.
4. `verify-workflows.mjs`'s header claims the script "asserts … that it walked at
   least one step"; it reports `usesSeen` but never asserts on it. The self-test
   covers the same risk more strongly, so this is a comment that overstates the
   code by one clause.

Merge it. The deploy is the real test, and this is the version of the change I
would want to be running when it happens.

